PR #128, opened 2024-09-18
🐛 Fix issue #125

### LINKED ISSUE #125: Focus not trapped in form
## Prerequisites

## Expected Behavior
Given that the event form is open
When the user TABs through all fields
Then the focus should restart at the beginning (the title field)
And when the user pressing ESC
Then the form should still close

## Current Behavior
After cycling through all form fields, the focus goes back to the URL

## Steps to Reproduce

1. Create event or draft

2. Hit TAB until focus goes to URL bar

Trap focus, either natively or using a library

## Context
Please include react-testing-library/jest tests to confirm your solution

DiegoMutre:
Now the focus is trapped in the EventForm and SomedayEventForm. This fixes #125 

A better solution can be implemented using the library [focus-trap-react](https://github.com/focus-trap/focus-trap-react). However, the `FocusTrap` component doesn't work with the `SomedayEventForm` component for some reason. I could try to figure out why, but first I'm proposing this solution to know if it's enough. The code implementation with focus-trap-react would be simpler and cleaner tho.

### DIFF
--- a/packages/web/src/common/utils/grid.util.test.ts
+++ b/packages/web/src/common/utils/grid.util.test.ts
@@ -1,0 +1,13 @@
+import { getLineClamp } from "./grid.util";
+
+describe("getLineClamp", () => {
+  it("uses a minimum value of 1", () => {
+    expect(getLineClamp(-1)).toBe(1);
+    expect(getLineClamp(0)).toBe(1);
+    expect(getLineClamp(1)).toBe(1);
+    expect(getLineClamp(1.818)).toBe(1);
+  });
+  it("uses value larger than 1 when possible", () => {
+    expect(getLineClamp(190.88)).toBeGreaterThan(1);
+  });
+});
--- a/packages/web/src/common/utils/grid.util.ts
+++ b/packages/web/src/common/utils/grid.util.ts
@@ -438,10 +438,11 @@ export const getAbsoluteLeftPosition = (
   return positionStart;
 };
 
-export const getLineClamp = (durationHours: number) => {
-  // how were these magic numbers determined?
-  const heightOfEvent = 54 * +durationHours.toFixed(2) || 0.25;
-  return Math.round((heightOfEvent - 7) / 16) || 1;
+export const getLineClamp = (height: number) => {
+  const min = 1;
+  const computed = Math.round((height - 7) / 16);
+  const lineClamp = Math.max(min, computed);
+  return lineClamp;
 };
 
 export const getPrevDayWidth = (today: Dayjs) => {
--- a/packages/web/src/views/Calendar/components/Event/Grid/GridEvent/GridEvent.tsx
+++ b/packages/web/src/views/Calendar/components/Event/Grid/GridEvent/GridEvent.tsx
@@ -1,9 +1,14 @@
-import React, { ForwardedRef, forwardRef, memo, MouseEvent } from "react";
+import React, {
+  ForwardedRef,
+  forwardRef,
+  memo,
+  MouseEvent,
+  useMemo,
+} from "react";
 import dayjs from "dayjs";
 import { Schema_GridEvent } from "@web/common/types/web.event.types";
 import { Measurements_Grid } from "@web/views/Calendar/hooks/grid/useGridLayout";
 import { WeekProps } from "@web/views/Calendar/hooks/useWeek";
-import { Text } from "@web/components/Text";
 import { Flex } from "@web/components/Flex";
 import {
   AlignItems,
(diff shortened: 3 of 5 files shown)


tyler-dane:
Hey @DiegoMutre, thanks for the PR!

Your `useFocusTrap` hook is clever. However, I agree that using `focus-trap-react` would likely be a more sustainable option for the following reasons:

1. React19 is deprecating `forwardRef`, which would require us to update this implementation anyway
2. Having to remember to add refs for the first and last element is prone to error, as some developers won't immediately understand why that's necessary
3. We already have an `onKeyDown` handler in that form, so having another one in a child element increases complexity

Can you say more about why `focus-trap-react` wasn't working with the `SomedayEventForm`, but it did with the regular `EventForm`? My inclination would be to refactor `SomedayEventForm` to make it compatible

DiegoMutre:
> Hey @DiegoMutre, thanks for the PR!
> 
> Your `useFocusTrap` hook is clever. However, I agree that using `focus-trap-react` would likely be a more sustainable option for the following reasons:
> 
> 1. React19 is deprecating `forwardRef`, which would require us to update this implementation anyway
> 2. Having to remember to add refs for the first and last element is prone to error, as some developers won't immediately understand why that's necessary
> 3. We already have an `onKeyDown` handler in that form, so having another one in a child element increases complexity
> 
> Can you say more about why `focus-trap-react` wasn't working with the `SomedayEventForm`, but it did with the regular `EventForm`? My inclination would be to refactor `SomedayEventForm` to make it compatible

Thanks for the feedback, @tyler-dane.

When wrapping the `StyledEventForm` in the `SomedayEventForm` component with the `FocusTrap`, it throws this error:

image

As I said, this only happens with the `SomedayEventForm`, and not with the `EventForm`. This is how the code looks like:

```
<FocusTrap>
      <StyledEventForm
        {...props}
        name="Some day Event Form"
        data-testid="somedayForm"
        id={ID_SIDEBAR_FORM}
        isOpen={true}
        onClick={stopPropagation}
        onKeyDown={onKeyDown}
        onMouseDown={stopPropagation}
        onMouseUp={(e) => {
          e.stopPropagation();
        }}
        priority={priority}
        role="form"
      >
        <StyledIconRow>
          <DeleteIcon onDelete={onDelete} title="Delete Someday Event" />
        </StyledIconRow>

        <StyledTitleField
          autoFocus
          name="Event Title"
          onChange={onChangeEventTextField("title")}
          placeholder="Title"
          role="textarea"
          title="title"
          value={title}
        />

        <PrioritySection
          onSetEventField={onSetEventField}
          priority={priority}
        />

        <RepeatSection
          bgColor={bgColor}
          onSetEventField={onSetEventField}
          recurrence={event.recurrence}
        />

        <StyledDescriptionField
          onChange={onChangeEventTextField("description")}
          placeholder="Description"
          value={event.description || ""}
        />

        <SaveSection priority={priority} onSubmit={_onSubmit} />
      </StyledEventForm>
</FocusTrap>
```

I will take a closer look at this and try to figure out how to make it work.

DiegoMutre:
Quick update:

- Clicking on the sidebar to create a new someday event **throws error**.
- Clicking on the "+" button to create a new someday event **throws error**.

- Clicking on a created someday event to open the event doesn't throw error.
- Clicking on the calendar grid to create a new event, either timed or all-day doesn't throw error.
- Clicking a grid event in the calendar, either timed or all-day doesn't throw error.

- Clicking a grid event and then clicking a someday event **throws error**.

In short, I've just discovered three instances where could throw the error displayed in the previous comment.

Btw, I discovered it last week but didn't have time to comment it. I'm still on it. I will fix it for this week.

tyler-dane:
@DiegoMutre Thanks for the update. Is this happening when rendering the `SomedayEventForm` or regular `EventForm`? If it's happening during both, that'd point to a flaw in how either the `focus-trap` library is being used or how compass code is written. 

If it just happens with `SomedayEventForm` but not `EventForm`, then I'd look into `SomedaysEventForm`'s parent components to find out the root cause

DiegoMutre:
> @DiegoMutre Thanks for the update. Is this happening when rendering the `SomedayEventForm` or regular `EventForm`? If it's happening during both, that'd point to a flaw in how either the `focus-trap` library is being used or how compass code is written.
> 
> If it just happens with `SomedayEventForm` but not `EventForm`, then I'd look into `SomedaysEventForm`'s parent components to find out the root cause

I found a better solution for the SomedayEventForm. I was suspicious of floating-ui because of the component FloatingPortal wrapping the SomedayEventForm so I wanted to know more about floating-ui, and I found out about the component [FloatingFocusManager](https://floating-ui.com/docs/FloatingFocusManager) which is like focus-trap but specially for floating-ui. I implemented it and works well. I don't know exactly why focus-trap doesn't work with the SomedayEventForm yet. But focus-trap is still used for the EventForm only.

It seems that everything works well, but I'm struggling with adding tests. For some reason the `test:web` script doesn't detect the new tests added; it keeps saying `No tests found, exiting with code 1`, even when I modify the existing tests.

tyler-dane:
Glad to hear that the floating-ui component worked -- nice find.

Do the tests you added work when run directly through `jest` on the CLI? Wonder if it's an issue with how I configured `yarn test:web` to run. 

Either way, feel free to push up what you have and I can give it a look. You can always rebase before merging if needed.

PR #176, opened 2024-12-08
♻️ Refactor: Removed ability to manually hide times from individual events

murilo9:
What I did:

- Removed the isTimesShown property from the data model and all other files, including tests (mocks, tests). All remaining tests have passed.
- Removed the CSS that renders the border around the times label upon hover.
- Removed the onClick handler on the event label.
- Refactored GridEvent's logic to not render the times label if event is past.


packages/web/src/views/Calendar/components/Event/Grid/GridEvent/Times/Times.tsx
tyler-dane, line 1: Can you remove as much of this as possible? Now that we're not letting the user toggle the times, we won't need `shouldRevealBox`, `StyledTimes`, `StyledTimesPlaceholder`.

In fact, even `Times.tsx` could probably be deleted, which would let us just move this to the `GridEvent`

```
      <Text role="textbox" size="xs" zIndex={ZIndex.LAYER_3}>
        {getTimesLabel(event.startDate, event.endDate)}
      </Text>
```

murilo9: Sure. Only a couple questions:
- Should times be displayed on drafts? If not, I'll remove the `isDraft` check.
- What is the purpose of the `isPlaceholder` check?

tyler-dane: Yes, times should be displayed in the event draft. That helps users understand their changes while dragging, resizing, and adjusting the times through the form inputs.

`isPlaceholder` tells the UI whether the event rectangle that is being displayed is the same as what the user is editing. Whenever a user edits a `GridDraft`, we display a new component (the `Draft`). The `GridDraft` can act as a placeholder to show the user the original event, which is helpful when dragging or resizing.

murilo9: Done

tyler-dane:
Looks good! 
I snuck in one change: always showing times when the event is a draft. After testing it out, I found this to be a nice UX improvement. Since a user is most likely to care about the times of a past event when they're actively editing it, I think this makes sense.

PR #181, opened 2024-12-10
🐛Fix: keep event draft during save process

### LINKED ISSUE #145: Keep new event on grid during save
## Prerequisites

## Expected Behavior
Event rectangle is always present. 

## Current Behavior
Event rectangle disappears from the screen temporarily while the event is being saved

## Steps to Reproduce

1. Optional: set network to slow 3G to increase delay

2. Click grid or all day row to create new event

3. Notice: event disappears from screen and then reappears 

This is apparent only during creating a new event, because there is no existing grid event to fall back on when the draft event is disappearing. There needs to be some way to either preserve the draft event on the screen or optimistically render the regular event without waiting for a successful response from the server. If going down the optimistic rendering route, be sure to handle how to revert the event if the save failed. 

## Context
As a result, users are less confident that their changes will be saved.

murilo9:
This fixes #145 

What I've done:

- Added code from this [POC](https://github.com/SwitchbackTech/compass/compare/main...poc/issue%23145).
- The POC's code managed to preserve the draft during save, but once I clicked on a different grid position to create a new event, it created a draft with the previous values and position.
- To fix the above issue, I changed `Draft.tsx:useEffect`'s logic to sync preservedDraft with draft when isProcessing is false.
- All automated tests have passed.

How I've tested this fix:

1. Clicked on a random position on the grid, and checked if event form opens.
2. Clicked outside the event form, and checked if event form closes, and the draft was gone.
3. CLicked on a new random position in the grid, and checked if event form opens.
4. Added event title and time, then clicked Save. CHecked that draft got preserved during save process.
5. Clicked on a new random position in the grid, and checked if a new draft was created in the correct position, with the event form propperly filled.


tyler-dane:
Hey @murilo9, I'm afraid the current state of this PR would do more harm than good due to the following issues I'm seeing

1. When creating a new timed event, the title flashes indefinitely. This doesn't happen every time, but often enough.
https://github.com/user-attachments/assets/b8c39a16-2201-4e34-80d9-e3bb0275af9f

2. The form now stays open until the frontend receives a response from the backend. Previously, the form closed immediately. This delay causes a user to question whether they clicked the Save button.

3. Clicking the empty space in the all-day row results in the event rectangle rendering, but not the form. Clicking the empty rectangle will then open the form. (This is occasional and may not be caused by this PR). 

----

Please spend a little time considering if there's a way we can change our approach to fix these issues without having to do a big refactor. If not, I'd rather put this issue on hold until after cleaning up our state management than introduce more bugs.

packages/web/src/views/Calendar/components/Event/Draft/Draft.tsx
```diff
@@ -40,21 +43,35 @@ export const Draft: FC<Props> = ({
...
+    }
+  }, [isProcessing, draft]);
+
+  if (isLoadingDOM) return null;
+  if (!preservedDraft) return null;
+
+  if (preservedDraft && !isDrafting && !isProcessing) {
+    return null;
+  }
+  if (!preservedDraft && !draft) return null;
```
tyler-dane, line 63: This effect is probably what's getting us into trouble.  But I'm wondering if `draft` is changing too frequently for this copied `preservedDraft` to stay in sync

tyler-dane (changes requested):
Hey @murilo9, I'm afraid the current state of this PR would do more harm than good due to the following issues I'm seeing

1. When creating a new timed event, the title flashes indefinitely. This doesn't happen every time, but often enough.
>https://github.com/user-attachments/assets/b8c39a16-2201-4e34-80d9-e3bb0275af9f
2. The form now stays open until the frontend receives a response from the backend. Previously, the form closed immediately. This delay causes a user to question whether they clicked the Save button.
3. Clicking the empty space in the all-day row results in the event rectangle rendering, but not the form. Clicking the empty rectangle will then open the form. (This is occasional and may not be caused by this PR).

Please spend a little time considering if there's a way we can change our approach to fix these issues without having to do a big refactor. If not, I'd rather put this issue on hold until after cleaning up our state management than introduce more bugs.

murilo9:
Regarding no. 1 and 3, I'll give a look on them. For no. 2, may I add a loading state to the Save button (disable the button and change its text to "Saving...")?

tyler-dane:
Adding a pending message ('Saving...') and disabling the button would mitigate confusion. But that's still a step backward compared to how it works now since we'd still be forcing the user to wait for the operation to finish. 

I'd rather focus on getting the event rectangle to always show without any strings attached. 

We could add some non-blocking pending state to show that the background operation is still processing as part of another enhancement.

murilo9:
I thikn I don't really understand the logic of `Draft.ts` yet, especially what makes it (re)render, the purpose of its variables (isDrafting, isProcessing, isLoadingDom, draft vs preservedDraft) and the effects of each combination of values for those variables. I'm afraid I'll have to give up on this issue for now and try another that I have a better chance to finish before the end of this week 😞

tyler-dane:
OK, appreciate the honesty. I'm going to close this so we can focus on other issues

murilo9:
May I retake this one?

PR #182, opened 2024-12-10
🐛 Fix: All day event doesn't disappear after changing the week

### LINKED ISSUE #178: All day event doesn't disappear after reassigning weeks
## Prerequisites

## Expected Behavior
Given that there is an all-day event
and the event's form is open
and a user changes the start and end dates
and those start and end dates are both in a new week
and the user clicks Save
then the all-day event rectangle should disappear from the screen
and the event should be visible when navigating to the correct week.

## Current Behavior
After saving the event, the rectangle still appears in the current week view.

This is happening for both new and existing events

## Steps to Reproduce

1. Click the all-day row to create an all-day event (or click an existing  all-day event)

2. Click the start date picker and assign the event to a new week. For example, if today is 12/9/2024, and the week in view is 12/8/24 - 12/14/24, then make the new start 12/17/24 (the next week)

3. Repeat for the end date

4. Click save
5. Notice: the event rectangle doesn't disappear
6. Refresh page
7. Notice: the event rectangle is gone now (due to re-fetching data during page load)

Consider adding a redux action that updates the draft state. Try to avoid adding lots of effects to the `DraftEvent`, which is already getting overly complex. 

Also remember to make the solution work for both new and existing all-day events.

## Context
This bug makes users less confident in Compass and it's ability to do the basic task of rescheduling events.

murilo9:
This fixes #178 

I realised the problem was the fact that it was assumed that _an event starting **and** ending in a different week would be an event that starts in a past week and ends in a future week_, wich is not necessarily true. That was making an event that, say, starts and ends in a past week to be rendered as an event that overlaps the current week, which is unitended.

To fix that, I've added a new `Category` to `event.types.ts`, called `NotAtThisWeek`, which is intended for events that starts **and** ends in a different week, but are _not_ `PastToFutureWeek`.

I also refactored the tests for `getAllDayEventWidth` to take into consideration the new logic of `pastToFutureWeek` and `notAtThisWeek`. All automated tests have passed.

**Obs**: this branch was created from my main branch, which was already merged with [fix/issue#145](https://github.com/murilo9/compass/tree/fix/issue%23145). That's why you'll see code unrelated to the current fix. I recommend you to approve this [PR](https://github.com/SwitchbackTech/compass/pull/181) first.


murilo9:
Temporarily closed due to not all tests have passed. I will refactor the tests related to this feature and reopen the PR again when done.

packages/web/src/__tests__/utils/grid.util/grid.util2.test.ts
```diff
@@ -137,9 +137,9 @@ describe("getAllDayEventWidth", () => {
 
   test("pastToFutureWeek", () => {
     const start = dayjs("2022-01-01");
-    const end = dayjs("2022-03-06");
-    const startOfWeek = dayjs("2022-03-12");
-    const endOfWeek = dayjs("2022-12-30");
+    const end = dayjs("2022-03-20");
+    const startOfWeek = dayjs("2022-03-06");
+    const endOfWeek = dayjs("2022-03-12");
```
tyler-dane, line 142: nice catch!

tyler-dane (changes requested):
I appreciate you coming up with an approach that works (I tested it). However, I do have some concerns

### Dependent Code
This code looks to be branched off #181. The same issues I voiced in that PR apply here. Whenever possible, please avoid making one pending PR dependent on another.

### The fix is resource-intensive and has a large code footprint

This implements the fix at the style-level

Here's what I mean by "style-level": 
An enum was added to the event category (`Category.NotAtThisWeek`). This category is used by the `getPosition()` util, which is used by the `GridEvent` and `AllDayEvent` to determine the height/width of the rectangle.

The shortcomings of this include:
1) The `NotAtThisWeek`isn't currently used in `getAllDayEventWidth`, so it just falls back to the default value anyway. 
2) It further entrenches us in the concept of a 'week', which will make it harder to refactor once we need to support other view durations (1 day, 4 day, 10 day, etc)
3) These position calculations are expensive, so running them only to determine that an event shouldn't be rendered is a waste of the user's compute/memory resources

**Suggestion**
Instead of fixing the issue at the util/style level, adjust the lifecycle of an event edit. This will let us remove the event automatically, without having to do additional calculations.

Here's how that'd look specifically

Update the submit handler to check whether an event should still be shown based on the latest data

```typescript
//useDraftUtil.ts
  const submit = (draft: Schema_GridEvent) => {
    const event = prepEvtBeforeSubmit(draft);

   // get this value by comparing the event's dates to 
   // the weekProps.component.startOfView and [...].endOfView
   const shouldRemove = true

    const _payload = {_id: event._id, event};
    const payload = shouldRemove ? {...payload, shouldRemove} : _payload
    const isExisting = event._id;
    if (isExisting) {
      dispatch(
        editEventSlice.actions.request(payload)
      );
      discard();
    } else {
      dispatch(createEventSlice.actions.request(event));
    }
  };
```

Add a step to the editEvent saga that reacts to the new shouldRemove directive

```typescript
//event.saga.ts


export function* editEvent({ payload }: Action_EditEvent) {
  try {
    yield put(eventsEntitiesSlice.actions.edit(payload));
    yield call(EventApi.edit, payload._id, payload.event, {
      applyTo: payload.applyTo,
    });

   if (payload.shouldRemove) {
       // remove the id from the week view Redux store:
      // events.getWeekEvents.value.data
   }

    yield put(editEventSlice.actions.success());
  } catch (error) {
    yield put(editEventSlice.actions.error());
    handleError(error as Error);
  }
}
``` 

As a result, any all-day event that changed weeks should be removed from the UI immediately after it's removed from the week view Redux slice. This helps us avoid managing that ourselves in util/styling code. 

Redux

murilo9:
Understood. I'll discard these changes and start over from a new branch then.

murilo9:
I'm stuck at the "remove the id from the week view Redux store" part. What do you mean by the "week view" store? I found a `getWeekEventsSlice`, but it doesn't contain any info about its state (afaik). It's also not clear for me how exactly that would impact the events displayed (or not) in the calendar.

tyler-dane:
> I'm stuck at the "remove the id from the week view Redux store" part. What do you mean by the "week view" store? I found a `getWeekEventsSlice`, but it doesn't contain any info about its state (afaik). It's also not clear for me how exactly that would impact the events displayed (or not) in the calendar.

Yes the `getWeekEvents` Redux slice is the one to look at. That's what I was showing in that screenshot above.

The ids in that slice are used to find the corresponding entity in the entity slice. If you remove an id from that week slice. Those ids are what the event selectors are using to decide which events are in view and which arent

PR #185, opened 2024-12-12
🐛 Fix: All day event doesn't disappear after changing the week - Attempt #2

murilo9:
This is a better, non-style-level fix for #178 

As you suggested, I've added some logic into `useDraftUtil.ts:submit` to check if the updated event will be rendered inside current view. If not, it marks the event to be removed from the store after the update. Now once we save an event with dates for a different week, it immediately disapers from the screen.

All automated tests have passed.

However, the "flicking" event with full width still persists for some milliseconds once you change weeks. I'm not sure yet what is the cause of this, but I will let the PR open while I give a look on this.

### DIFF
--- a/packages/web/src/ducks/events/event.types.ts
+++ b/packages/web/src/ducks/events/event.types.ts
@@ -77,6 +77,7 @@ export interface Payload_EditEvent {
   applyTo?: Categories_Recur;
   _id: string;
   event: Schema_Event;
+  shouldRemove: boolean;
 }
 
 export interface Payload_GetPaginatedEvents extends Filters_Pagination {
--- a/packages/web/src/ducks/events/event.types.ts
+++ b/packages/web/src/ducks/events/event.types.ts
@@ -74,10 +74,10 @@ interface Payload_DeleteEvent {
 }
 
 export interface Payload_EditEvent {
-  applyTo?: Categories_Recur;
   _id: string;
   event: Schema_Event;
-  shouldRemove: boolean;
+  applyTo?: Categories_Recur;
+  shouldRemove?: boolean;
 }
 
 export interface Payload_GetPaginatedEvents extends Filters_Pagination {
--- a/packages/web/src/ducks/events/sagas/event.sagas.ts
+++ b/packages/web/src/ducks/events/sagas/event.sagas.ts
@@ -166,6 +166,9 @@ export function* deleteSomedayEvent({ payload }: Action_DeleteEvent) {
 export function* editEvent({ payload }: Action_EditEvent) {
   try {
     yield put(eventsEntitiesSlice.actions.edit(payload));
+    if (payload.shouldRemove) {
+      yield put(eventsEntitiesSlice.actions.delete({ _id: payload._id }));
+    }
     yield call(EventApi.edit, payload._id, payload.event, {
       applyTo: payload.applyTo,
     });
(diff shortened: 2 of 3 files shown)


packages/web/src/ducks/events/event.types.ts
```diff
@@ -77,6 +77,7 @@ export interface Payload_EditEvent {
   applyTo?: Categories_Recur;
   _id: string;
   event: Schema_Event;
+  shouldRemove: boolean;
```
tyler-dane: Making this a required property without updating the call in `useSidebarUtil` is going to cause issues. 

Let's make this optional, as that gives us more flexibility without bloating the payload in other areas of the app like the sidebar (which doesn't have the same use-case of needing an extra property to check for removal)

As you're seeing, our tests are lacking, especially on the frontend. We can't rely on our test suite passing to ensure that we're not introducing more bugs

tyler-dane: Made this optional in my recent commit

tyler-dane (commented):
This is much better so far.

Would like more clarification on the flickering you mentioned before QA-ing further.

packages/web/src/ducks/events/sagas/event.sagas.ts
```diff
@@ -164,10 +164,15 @@ export function* deleteSomedayEvent({ payload }: Action_DeleteEvent) {
...
-    yield put(eventsEntitiesSlice.actions.edit(payload));
-    yield call(EventApi.edit, payload._id, payload.event, {
-      applyTo: payload.applyTo,
+    shouldRemove
+      ? yield put(eventsEntitiesSlice.actions.delete({ _id }))
+      : yield put(eventsEntitiesSlice.actions.edit(payload));
+
+    yield call(EventApi.edit, _id, event, {
+      applyTo: applyTo,
     });
```
tyler-dane, line 176: I'm using the `shouldRemove` property to determine whether to call `.edit` or `.delete` (previously they were both called sequentially). 


Technically, we should also be removing the id from the `getWeekEventsSlice` if `shouldRemove` is true. That is so the week slice accurately depicts what is in view. However, I found that to cause the event to stay on the screen. Instead of going down that rabbit hole, I'm OK introducing this as a code smell. In the long term, I'd like to replace Redux Saga (and possibly Redux) with something less verbose and complex. So, I don't want to invest a lot more into getting everything with Redux and Redux Saga to work perfectly.

murilo9: Imo keeping redux is fine for global state, maybe we could just simplify the store a bit, and let the requests and async work outside the store, or use Redux Thunks for that.

murilo9:
How to reproduce the "flickering" event bug:
1. Add an all-day event in a different week.
2. Move the view to that week, so you can see the all-day event. It should render properly.
3. Move back to the previous week, you'll note that an all-day event with the duration of the whole view quickly shows (for less than a second) then disappears.

I'm not sure what is causing that but I suspect it might be something to do with the order of the steps in `editEvent` saga.

tyler-dane:
Thanks for including the detailed steps to reproduce it. Since that behavior exists in `main` now, we don't need to block this PR. I create a new bug to address it in the future: #186 

Doing a little more QA on this one, but it's in pretty good shape. Please feel free to continue working on your other issue while I continue testing/tweaking

Issue #165, opened 2024-11-14
Make header labels more accurate


tyler-dane:
The current, somewhat-confusing labels:
![Image](https://github.com/user-attachments/assets/69546913-0675-46be-8809-b69d0b5f8f66)

Rules
1. Display the month when viewing a week that is entirely in the current year
2. Display the the month+year when viewing a previous/future year
3. When viewing a week that spans multiple years, display the label in the format of `{Short Month} {Short Year} - {Short Month} {Short year}`

## Acceptance Criteria

- [ ] All labels follow the above rules
- [ ] Unit test(s) provided that validate each rule

Here is a reference table for all scenarios:

| Today    | Week In View | Heading label | Someday Week label | Someday Month label |
| -------- | ------- | ---- | ---- |---- |
| 1/8/25  | 1/5/25 - 1/11/25    | January | 1.5-11 | January |
| 12/30/24  | 1/5/25 - 1/11/25    | January 2025 | 1.5-11 | January |
| 12/30/24  | 12/29/24 - 1/4/25    | Dec 24 - Jan 25 | 12.29 - 1.4 | December |

## Context
Since Compass is a minimal calendar, this should help us minimize the text on the screen while also making it clear what dates the user is currently viewing

## Implementation Guidance
To confirm how it looks in the UI, you can hard-code the `today` value to in `useToday.ts` like so:
```typescript
const today = dayjs("2024-01-04");
```

murilo9:
I'll be working on this one.

PR #188, opened 2024-12-14
♻️ refactor: make header labels more accurate

murilo9:
This resolves issue #165 

What I've done:

- Created a `getCalendarHeadingLabel` util function for `Header.tsx`.
- Refactored `getWeekRangeLabel` util function so it can return two different formats.
- Added a `web.date.util.test.ts` file, containing unit tests for `getWeekRangeLabel` and `getCalendarHeadingLabel`.
- Left `getMonthListLabel` unmodified, as it already behaves as expected (returns the full month name of week start).

I didn't write unit tests for `getMonthListLabel` due to its simplicity, but I can add it if you see it as necessary.
All unit tests (including the added ones) have passed.

### DIFF
--- a/packages/web/src/common/utils/web.date.util.test.ts
+++ b/packages/web/src/common/utils/web.date.util.test.ts
@@ -1,0 +1,61 @@
+import dayjs from "dayjs";
+import { getCalendarHeadingLabel, getWeekRangeLabel } from "./web.date.util";
+
+describe("getWeekRangeLabel", () => {
+  it("should return 'M.D - D' format when week is within single month", () => {
+    const weekInViewStart = dayjs("2025-01-05");
+    const weekInViewEnd = dayjs("2025-01-11");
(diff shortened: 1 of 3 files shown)


murilo9:
Some observations: in my opinion, the format of Someday Week label could be improved, as "M.D - D" is not very intuitive. Maybe the "MMM D - D" format could be better? That way it would be more intuitive and not repeat the full month name ("MMMM") like in the Someday Month label.

tyler-dane:
> Some observations: in my opinion, the format of Someday Week label could be improved, as "M.D - D" is not very intuitive. Maybe the "MMM D - D" format could be better? That way it would be more intuitive and not repeat the full month name ("MMMM") like in the Someday Month label.

Thank you for the suggestion! I'll keep this in mind and compare it to other user feedback on this change

tyler-dane (approved):

PR #194, opened 2024-12-18
🐛 fix: Optional email service proper handling

that-one-arab:
### Description:
As per the docs, email service should be optionally provided.

But the code requires it.

If we do not provide it, an exception is raised due to missing environment variables

If we provide it with incorrect values (such as the values in the example [.env](https://docs.compasscalendar.com/docs/get-started/configuration/env) we get error `TypeError: Converting circular structure to JSON` when we try to login using google oauth.

This error, when traced, is due to the below request error sent to `api.convertkit.com`
```
data: { error: 'Authorization Failed', message: 'API Key not valid' }
```

This is how the request URL looks like `https://api.convertkit.com/v3/tags/YOUR_LIST_ID/subscribe?api_secret=UNIQUE_SECRET_FROM_YOUR_CONVERTKIT_ACCOUNT&email=my-email@gmail.com&first_name=my-name` so an error is expected to happen here.

This PR handles the aforementioned case

### Adding Feature(s):
- Make email service optional.

### Fixing Issue(s):
- Error during oauth sign in due to incorrect email environment variables

### DIFF
--- a/CODE_OF_CONDUCT.md
+++ b/CODE_OF_CONDUCT.md
@@ -1,4 +1,4 @@
 # Code of Conduct 🤝
 
-The Compass Code of Conduct is available on the official doc:
-[https://docs.compasscalendar.com/docs/code-of-conduct](https://docs.compasscalendar.com/docs/code-of-conduct)
+The Compass Code of Conduct is available here:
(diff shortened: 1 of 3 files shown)


tyler-dane:
Thanks for being proactive about creating a solution for this.

I'd like to get away from using local variables like `_error` and `_optional` and instead use a validation library like `zod` to do this for us. Please leave this as-is for now. I will follow up with a follow-up commit that does that.

that-one-arab:
reopend here

PR #197, opened 2024-12-22
🐛 fix: keep new event on grid during save

that-one-arab:
Fixes [this](https://github.com/SwitchbackTech/compass/issues/145)

Uses an optimistic UI rendering approach.

I believe Optimistic UI approach is the simplest and easiest solution under the current circumstances.

There are a couple of edge cases that needs to be handled that I believed were out of scope of this issue, but should be tackled in the future; during the timespan where the optimistically generated event is created and the HTTP request returns a response from the server:

- The event cannot be drag and dropped
- The event cannot be deleted (would delete a local version but the API would re-generate a new one after the HTTP request returns a response)

We can tackle these separately

### DIFF
--- a/packages/core/src/types/event.types.ts
+++ b/packages/core/src/types/event.types.ts
@@ -56,6 +56,7 @@ export interface Schema_Event {
   title?: string;
   updatedAt?: Date;
   user?: string;
+  optimistic?: boolean;
 }
 
 export interface Query_Event extends Query {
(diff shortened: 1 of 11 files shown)


tyler-dane (commented):
I'm gonna do a little more cleanup and QA, but this is in pretty good shape. 

Thanks, @that-one-arab!

that-one-arab:
> I'm gonna do a little more cleanup and QA, but this is in pretty good shape.
> 
> Thanks, @that-one-arab!

Thanks @tyler-dane !

I'll keep an eye on the cleanups you're doing, learn new things in the process.

that-one-arab:
@tyler-dane I added additional commits that continues from where you left off

tyler-dane:
Alright, deployed and tested. Congrats on your first PR! 🙌

Issue #190, opened 2024-12-14
Event saved with wrong time when drag and dropping from sidebar


tyler-dane:
## Prerequisites

- [x] Using an up-to-date main branch

## Expected Behavior
GIVEN that a user has their mouse pressed and positioned over the grid
AND the preview event rectangle is rendered at the correct position
AND the preview event shows the desired correct time (eg `10 am`) 
AND the user releases the mouse
THEN an event rectangle is immediately displayed on the grid
AND the event time is the same as when the mouse was released

 
## Current Behavior
If a user continues to move their mouse after they drop and between the time when it was saved on the backend, then the start and end dates will be based on their most recent mouse position -- not their original. 

Notice: I release the mouse on Friday, then move over to Sunday. The event is saved as Tuesday 🙈

https://github.com/user-attachments/assets/48e9c7e4-5ab0-4de2-97a7-c98c19b52de3

## Steps to Reproduce

1. Create a sidebar event

2. Click the event and drag it over the grid

3. Note the day and time of the event

4. Release your mouse, while quickly moving it to the left or right
5. Notice: the event is saved at the latest mouse position, not the mouse position when you released it

## Possible Solution (Not obligatory)

We need a way to accurately capture the desired date/time at the time of mousing up, rather than relying on the mouse position's X/Y coordinates, which can change during the API request's lifecycle. 

## Context
This makes users frustrated, because what is saved in the UI doesn't match their intentions.

that-one-arab:
Assign me this please

tyler-dane:
> Assign me this please

Assigned!

PR #198, opened 2024-12-24
🐛 fix(web): Event saved with wrong time when drag and dropping from sidebar

that-one-arab:
Fixes [issue](https://github.com/SwitchbackTech/compass/issues/190)

This appears to be a react-beautiful-dnd issue due to the `onDragEnd` event not being called immediately on mouseup

Upon further investigation the root cause was due to its drop animation behavior causing a delay to the `onDragEnd` event

For more info, see: https://github.com/atlassian/react-beautiful-dnd/blob/master/docs/guides/drop-animation.md#skipping-the-drop-animation

Right now, the event renders with the correct date and time, but only renders after the http request returns a response

I handled optimistically inserting the to-be-rendered event, which is why I rebased this branch off of `that-one-arab:bug_keep_new_event_on_grid_during_save` in [here](https://github.com/SwitchbackTech/compass/pull/197) since it has some core optimistic insertion related logic, but due to us making refactors in that branch I needed to refactor the logic in this branch too but ran out of time, so I could not commit this enhancement.

Regardless, the main issue is tackled and hopefully we get around doing the following, in the same PR or a different PR, sometime in the future

- optimistically inserting the grid event
- optimistically removing the someday event

### DIFF
--- a/packages/web/src/views/Calendar/components/Sidebar/SomedayTab/SomedayEvents/SomedayEvent.tsx
+++ b/packages/web/src/views/Calendar/components/Sidebar/SomedayTab/SomedayEvents/SomedayEvent.tsx
@@ -1,6 +1,10 @@
 import { Key } from "ts-key-enum";
 import React, { Dispatch, SetStateAction, useEffect, useState } from "react";
-import { DraggableProvided } from "@hello-pangea/dnd";
+import {
+  DraggableProvided,
+  DraggableStateSnapshot,
+  DraggableStyle,
+} from "@hello-pangea/dnd";
 import { FloatingFocusManager, FloatingPortal } from "@floating-ui/react";
 import { Categories_Event, Schema_Event } from "@core/types/event.types";
 import { Schema_GridEvent } from "@web/common/types/web.event.types";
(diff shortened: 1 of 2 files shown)


that-one-arab:
@tyler-dane Git history fixed. Ready for review

tyler-dane:
Created #203 to track the follow-up enhancement of optimistic rendering for DND events

tyler-dane:
This approach of simply disabling the animation come with the benefit of being a small & contained change. I'm a little concerned that it's a workaround, based on the quote in that issue:

> Generally speaking you should be avoiding this. A drop animation is an important affordance to communicate placement. Our drop animations do not prevent the user from dragging something else while the animation is running.

> If you are seeing a strange drop behaviour, such as dropping to the wrong spot, our recommendation is to raise an issue as it could be a bug with react-beautiful-dnd or a setup issue.

I suspect the root issue is how we're using `react-beautiful-dnd`, rather than it being a bug. That's because we're not seeing any drop animation to begin with. It's nice that disabling it happens to fix this linked bug, but I think we're still going to run into DND issues going forward, and this might complicated fixing the root cause.

BUT .... I'm OK with that for now. Once the core features are more stable, I'd like to consider moving to a more modern library that is getting more maintenance, like [dnd kit](https://dndkit.com/). If we're going to migrate DND libraries, it doesn't make much sense to over-invest in getting this one to work for all cases. 

Nice job finding a quick solution

tyler-dane (approved):

that-one-arab:
> This approach of simply disabling the animation come with the benefit of being a small & contained change. I'm a little concerned that it's a workaround, based on the quote in that issue:
> 
> > Generally speaking you should be avoiding this. A drop animation is an important affordance to communicate placement. Our drop animations do not prevent the user from dragging something else while the animation is running.
> 
> > If you are seeing a strange drop behaviour, such as dropping to the wrong spot, our recommendation is to raise an issue as it could be a bug with react-beautiful-dnd or a setup issue.
> 
> I suspect the root issue is how we're using `react-beautiful-dnd`, rather than it being a bug in their library. That's because we're not seeing any drop animation to begin with. It's nice that disabling it happens to fix the linked Compass bug, but I think we'll still run into DND issues as we advance, which might complicate fixing the root cause.
> 
> BUT .... I'm OK with that for now. Once the core features are more stable, I'd like to consider moving to a more modern library that is getting more maintenance, like [dnd kit](https://dndkit.com/). If we're going to migrate DND libraries, it doesn't make much sense to over-invest in getting this one to work for all cases.
> 
> Nice job finding a quick solution

Thanks for your input

It is as you said, the issue is not with the library. The reason why we're not seeing the drop animation is due to us setting the someday event's opacity to 0 when `isOverGrid` and `isDragging` are true, and obviously we need this to happen to handle some UX related logic.

I just wanted to shed more light on the root issue. Since this is expected behaviour within their library and our own usage seems fine this seemed like the best workaround, but I'm confident this won't creep and cause a larger mess.

Issue #201, opened 2024-12-25
Event can still be changed during optimistic render


tyler-dane:
### Where did this happen?

Local code with an up-to-date main branch

### Expected Behavior

An event shouldn't be allowed to be edited or deleted while it is being rendered optimistically

### Current Behavior

During the timespan where the optimistically generated event is created and the HTTP request returns a response from the server:

The event cannot be drag and dropped
The event cannot be deleted (would delete a local version but the API would re-generate a new one after the HTTP request returns a response)

### Steps to Reproduce

1. Create Event
2. Change event
3. While event change is being processed on backend, change the event again

### Possible Solution

Prevent the changes while an event is optimistically rendered

### Context

#197 introduced optimistic rendering.

that-one-arab:
Assign me this please

PR #206, opened 2024-12-28
🐛 Fix: Disable newly created event from being editing during optimistic render cycle

that-one-arab:
### Description

While you already proposed a solution for this issue [here](https://github.com/SwitchbackTech/compass/issues/201#issue-2758833319), preventing the user from initially DnD or editing the event did not seem very UI/UX friendly to me.

I thought about and researched other solutions, the best solution to handle this issue is to implement **Transactional State Management** to handle this type of transient data, but this implementation itself is split into many strategies.

Each strategy has its own pros and cons, but they all share 1 same con and that is they are difficult to implement and adds lots of overhead.

The 2nd best approach was to simply display a loading indication when the event is initially being created, removing the optimistic approach entirely.

But this would:

- Undo some of the work we did
- Conflict with a comment you gave [here](https://github.com/SwitchbackTech/compass/pull/181#issuecomment-2538639603)

So the best thing we can do here is to implement your proposed solution which is disabling edits during its optimistic status.

I tried my best to make it UI/UX friendly by displaying a different cursor and a distinct background color to indicate we are doing some loading to the user.

I included all this extra context as logs for future reference and your own consideration.

### Demo
[Peek 2024-12-28 16-31.webm](https://github.com/user-attachments/assets/4dad0455-0cb7-432e-876e-7bab70a11999)

### DIFF
--- a/packages/web/src/common/styles/theme.util.ts
+++ b/packages/web/src/common/styles/theme.util.ts
@@ -30,6 +30,13 @@ export const hoverColorByPriority = {
   [Priorities.SELF]: brighten(c.blueGray200),
 };
 
+export const disabledColorByPriority = {
+  [Priorities.UNASSIGNED]: darken(c.blueGray400, 30),
+  [Priorities.WORK]: darken(c.blueGray100, 30),
+  [Priorities.RELATIONS]: darken(c.teal, 30),
+  [Priorities.SELF]: darken(c.blueGray200, 30),
+};
+
 export const gradientByPriority = {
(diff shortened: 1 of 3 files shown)


tyler-dane:
Thank you for sharing your thought process, @that-one-arab. I like how your solution is simple yet also offers an intuitive UX. I especially like how the styles only change onHover, allowing users to not have to think about what's happening behind the scenes. 

I followed up with one cleanup commit.  The original implementation worked fine, but it came with the overhead of another color map in theme.util. We're going to get away from these as we support more priorities, so this method lets us scale with more ease

PR #208, opened 2024-12-31
🐛 Fix: zod `EMAILER_` validation to handle empty strings

that-one-arab:
I ran the project only to get a zod error that my `EMAILER_` fields need to contain at least 1 character. Weirdly enough the `EMAILER_` env variables are optional.

This comes back to `zod`'s way of handling empty strings, based on this comment from the creator of `zod` [here](https://github.com/colinhacks/zod/issues/310#issuecomment-794533682)

This PR implements the suggestion in the comment

This should prevent developer confusion when they have `EMAILER_` env variables represented as empty strings not working when running locally, which should accelerate their onboarding process.


tyler-dane:
Thank you for taking the initiative to fix something you found frustrating as a dev. 

However, I question whether others will run into the same issue, as we now have an `.env.example` file that shows the expected format of this value. I imagine they'd understand to simply change the placeholder value in place of the actual value rather than fall back to an empty string.

Accepting an empty string along with `undefined` for these values would help us avoid startup errors, but it would also open us up to more bugs from the overhead of supporting multiple types. 

The general goal of the parsing in `env.constants` is to parse the `.env`, validate the inputs, and return a safe version object, where each property has one type. That lets developers have the confidence to work with environment values, rather than having to constantly think about the edge cases that might arise when supporting both `undefined` and `""`

I'm going to close this for the above reasons. But if anyone else runs into this issue, we can link it to this PR and revisit.

Issue #203, opened 2024-12-26
Optimistically render when drag-and-dropping event


tyler-dane:
### Feature Description

When drag-and-dropping a sidebar event onto the grid, the event currently disappears briefly while the request is made to the backend, and then re-appears after receiving the backend response. This experience is not great UX, as it leaves the user questioning if their change was lost or errored out. 

Instead, immediately render the event on the grid and remove the event from the sidebar. 

Follow up work from #197, #201, and #190

### Use Case

This'll save users the time and brief stress worrying about whether their DND action worked or not

### Additional Context

https://github.com/user-attachments/assets/aba5e896-bc28-47b1-8a4d-7ae4823de96d

that-one-arab:
Assign me this please

Issue #207, opened 2024-12-29
Site is refreshing after resizing or dragging event


tyler-dane:
### Where did this happen?

Production app, Local code with an up-to-date main branch

### Expected Behavior

Resizing or dragging an event shouldn't trigger a page refresh

### Current Behavior

Full page refresh happens sporadically after resizing or dragging an existing event

https://github.com/user-attachments/assets/ecbfeb41-89a6-4f15-8d53-6b83e42cf97f
> Notice how the page refreshes automatically after I simply click the event after dragging it.

**UI**
The UI automatically refreshes as part of the error handling for our custom 601 error, so it's hard to show how this looks beyond the above video.

You can disable that refreshing if you'd like to see the state of the UI before the refresh happens

**Backend**
This is what you'll see in the backend logs

<img width="1185" alt="Screen Shot 2024-12-31 at 6 17 48 PM" src="https://github.com/user-attachments/assets/605f7361-333c-4bd1-bf5c-a44a5fbbd172" />

**DB**
If you look in the mongo DB, you'll see the event is missing a `user` property

### Steps to Reproduce

1. create new event
2. drag/resize it
3. continue until the page is refreshed
4. view the logs and you'll see the following backend error: 

```
app:error.handler: {"result":"Prompt Redux refresh","description":"Invalid event id (most likely)","statusCode":601,"isOperational":true}
```

### Possible Solution

This is caused by the request event payload missing the `user` value. This `user` is used by the backend in `event.queries:updateEvent()` to identify the event to update. 

Because the `user` value is missing, the event isn't found, which triggers the backend to throw the following error

```
 throw error(EventError.NoMatchingEvent, "Prompt Redux refresh");
```

The frontend obliges, by refreshing the entire page in an attempt to resync the redux store. (In fact, this isn't necessary, as the Redux store likely wasn't the problem)

### Context

This is a very intrusive interruption that waste's a user's time and makes them have less confidence in the app.

tyler-dane:
@that-one-arab can you pick this up?

that-one-arab:
Yup leave it to me

tyler-dane:
@that-one-arab I don't have a video handy, but added a few more details

omarNaifer12:
Hi,
I’m trying to contribute, and the documentation says I need a Google Cloud account to get OAuth credentials. However, when I try to create an account, it requires a credit card, which I don’t have. Is there any other solution to contribute, or is this the only way?

[...]

PR #209, opened 2024-12-31
🐛 Fix: Optimistically render when drag-and-dropping event

that-one-arab:
Fixes [issue](https://github.com/SwitchbackTech/compass/issues/203)

In this commit, whenever we drag and drop a someday event to the grid, we now handle:
- Optimistically removing the someday event from the sidebar
- Optimistically inserting the grid event

We also error handle the optimistic flow and revert if needed

[Peek 2024-12-31 07-50.webm](https://github.com/user-attachments/assets/4b079052-ceb1-4bd8-bcbe-20e027def826)


tyler-dane (commented):
This approach is looking good. I started a little cleanup to make the saga easier to understand. I'll continue the cleanup and merge after QAing a little more.

Can you plz start on #207 next?

that-one-arab:
@tyler-dane Yep I agree. I'll try to make more of these refactors as I understand the codebase more and more.

Issue #200, opened 2024-12-24
Pre-emptively trigger login after google token expired


tyler-dane:
### Feature Title

Pre-emptively trigger login after google token expired

### Feature Description

Currently, the user's session might be active, but their google access token has expired. Compass doesn't know that the google access token expired, though, because it doesn't proactively check for it. Instead, it tries to make a normal event change request. Then google responds letting us know that the token expired, which we use to log the user out and redirect them to the sign in page.

The down side of this is that the user thought they were all good and spent time making/updating an event, only to be promptly redirected to login without any warning.

Instead, we should redirect the user before they're able to make any changes, providing a nice UX/message about why this is important.

### implementation idea(s):
- use the WebSocket's connection as a trigger to run a preliminary check
- if a user returns the compass after a long time, the session with be refreshed. before/after that happens, do a preemptive GET request on a Google calendar resource. This will let us know if their access token is still valid. There's probably a better way to

tyler-dane:
@that-one-arab Can you comment in this issue? I know you're working on it, but GitHub doesn't let me assign it to you until you interact in the comments

that-one-arab:
@tyler-dane lol thats funny, yes sure here's my comment

PR #212, opened 2025-01-03
🐛 Fix: Pre-emptively trigger login after google token expired

that-one-arab:
Fixes [issue](https://github.com/SwitchbackTech/compass/issues/200)

This pull request includes multiple changes to the authentication system, focusing on verifying Google tokens, enhancing error handling, and improving the user authentication flow. The most important changes include adding a new method for verifying Google tokens, updating the Google authentication service, and modifying the frontend authentication checks.

### Backend Changes:
* **Google Token Verification:**
  - Added `verifyGToken` method in `auth.controller.ts` to check if a user's Google access token is still valid.
  - Updated `auth.routes.config.ts` to include a new route for verifying Google tokens. [[1]](diffhunk://#diff-cbfb2cecf9bac48b7c95bb2ca2e12d204d65417b780485ba7854a842117d0404R8-R36) [[2]](diffhunk://#diff-cbfb2cecf9bac48b7c95bb2ca2e12d204d65417b780485ba7854a842117d0404L41-R50)
  - Introduced `Result_VerifyGToken` interface in `auth.types.ts` to handle the response from Google token verification.

* **Google Authentication Service Updates:**
  - Added `getGAuthClientForUser` method in `google.auth.service.ts` to obtain Google authentication clients for users. [[1]](diffhunk://#diff-2358f4b232a29c55e7c7cfcc5be62d708bf4489d599230003f648b2ccdfb4322R6-R58) [[2]](diffhunk://#diff-2358f4b232a29c55e7c7cfcc5be62d708bf4489d599230003f648b2ccdfb4322L27-R69)
  - Implemented `getAccessToken` method in `GoogleAuthService` to fetch access tokens and handle errors.
  - Introduced `NoGAuthAccessToken` error in `error.constants.ts` for handling missing Google access tokens.

### Frontend Changes:
* **Authentication Flow Enhancements:**
  - Refactored `ProtectedRoute` to use `useAuthCheck` for verifying both user and Google sessions.
  - Created `useAuthCheck` hook to manage authentication checks and session validation.
  - Updated `Login.tsx` to handle authentication state and navigate accordingly. [[1]](diffhunk://#diff-74ad512bf21ae2737da2fd7b1866f33e1fc08f0b7053731993a1a142a75fb9a3L3-R10) [[2]](diffhunk://#diff-74ad512bf21ae2737da2fd7b1866f33e1fc08f0b7053731993a1a142a75fb9a3R24-R37) [[3]](diffhunk://#diff-74ad512bf21ae2737da2fd7b1866f33e1fc08f0b7053731993a1a142a75fb9a3R51-R58) [[4]](diffhunk://#diff-74ad512bf21ae2737da2fd7b1866f33e1fc08f0b7053731993a1a142a75fb9a3L88-L90) [[5]](diffhunk://#diff-74ad512bf21ae2737da2fd7b1866f33e1fc08f0b7053731993a1a142a75fb9a3L111-L117)

* **API Updates:**
  - Added `validateGoogleAccessToken` method in `auth.api.ts` to validate Google tokens via the backend. [[1]](diffhunk://#diff-cc8a2b17d2a3737a8dbe1dbcce13e860d33703df354ffccefdef14cd663470e8L1-R4) [[2]](diffhunk://#diff-cc8a2b17d2a3737a8dbe1dbcce13e860d33703df354ffccefdef14cd663470e8R13-R25)

These changes collectively improve the robustness of the authentication system by ensuring that Google tokens are valid and handling authentication states more effectively on the frontend.

----

## How to Test
In lieu of of integration tests, here are the steps taken during manual QA to validate functionality.

Clear cache & cookies or do the following tests in a incognito browser

**Auth Check**
This is done in `ProtectedRoute`, which is triggered on page loads. Validate this by doing a full-page refresh, loading a new tab, or going to a different website in your current tab and then going back to Compass. You'll notice GET requests to `/api/auth/google`, which indicates we're checking the access token validity. 

Note: ensure these checks only occurs on page refreshes. We don't want to run those excessively.

**Search Params**
This PR adds a URL search param when an redirecting to login after a session expired (either user or google access token), which is intended to help the UI display more information to the user about why they need to re-login. 

Instead of waiting until your google access token expires, you can simulate that scenario by:
1. Running backend and web in dev
2. Signing in like normal
3. Update `getAccessToken` to always throw (comment out the `if(!token)` part. 
4. Wait until the backend finishes hot reloading
5. Reload the Compass page or open a new tab and go to Compass (this will trigger the auth check from `ProtectedRoute`)
6. Notice: you'll be redirected to `/login` and the URL will include a param called: `reason=gauth-session-expired` 

**Login redirect**
Auto-redirecting to login is a nice UX. However, it opened up a bug where a user who is already logged in and then goes to `/login` gets caught in an infinite loop of redirects and API calls. To address that, we're allowing the user to see the `/login` page regardless of auth status. If a user is already authenticated and then clicks 'Sign in with Google', however, they will simply be redirected to the root page rather than going through the auth flow again. 

You can test this scenario by logging in, going to `/login`, and clicking that button, and then looking at the Network tab and backend logs. You'll notice that no auth API requests are being made, which is expected.


packages/backend/src/auth/controllers/auth.controller.ts
```diff
@@ -63,6 +66,33 @@ class AuthController {
...
+    } catch (error) {
+      if (error instanceof BaseError && error.result === "No access token") {
+        res.promise({ valid: false, error: "Invalid Google Token" });
+      } else {
+        res.promise({
+          valid: null, // We don't know if the session is valid or not, so we return null
+          error,
+        });
+      }
+    }
```
that-one-arab: I would've liked to handle this error more consistently with how other google related errors are handled but I couldn't establish a pattern in functions like `isInvalidGoogleToken`

packages/backend/src/auth/services/google.auth.service.ts
```diff
@@ -3,17 +3,56 @@ import { OAuth2Client, TokenPayload } from "google-auth-library";
...
 import { error } from "@backend/common/errors/handlers/error.handler";
+import { Schema_User } from "@core/types/user.types";
+import { WithId } from "mongodb";
 
 import compassAuthService from "./compass.auth.service";
 
 const logger = Logger("app:google.auth.service");
 
-export const getGcalClient = async (userId: string) => {
+export const getGAuthClientForUser = async (
```
that-one-arab, line 22: Extracted as its own function for reusability. 
Gave the option of providing a mongodb user object or just the userId
Providing the user object directly will save a db call for better performance

packages/web/src/views/Login/Login.tsx
```diff
@@ -20,7 +23,18 @@ import {
...
 } from "./styled";
 
+const clearSession = async () => {
+  try {
+    await SyncApi.stopWatches();
+    await signOut();
+  } catch (error) {
+    console.error("Failed to clear session", error);
+  }
+};
```
that-one-arab: We need to do this to ensure that no leftover session data exists, could cause issues with determining reason for not being authenticated (since we authenticate with both Compass API and Google OAuth API)

tyler-dane:
No big issues with this at first glance. I want to QA this one more thoroughly to make sure, though. 

I'm going to hold off on doing that while I wrap up #203 and #196. 

In the meantime, please find an issue from the 'Ready' column of [the board](https://github.com/orgs/SwitchbackTech/projects/4/views/2) to work on. Alternatively, you could find an issue in the Backlog to refine. By "refine", I mean thinking it through more, adding details, and updating the issue with a high-level description of your suggested implementation. After doing that I'll read through it and give it the 👍 if it looks good to start working on.
 
 Since you're more familiar with the codebase and have been producing good work, this could be a good time to pick up a more difficult or larger issue if you're up for it.

that-one-arab:
@tyler-dane Thank you! I think I'll opt to work on issues based on the priorities you assigned them since we have a lot of them as ready. Once we start running out of ready issues I'll work on refining existing backlog items

packages/web/src/auth/ProtectedRoute.tsx
```diff
@@ -3,17 +3,29 @@ import Session from "supertokens-auth-react/recipe/session";
...
+
+      const isAuthenticated = isSessionValid && isGAuthSessionValid;
       setIsAuthenticated(isAuthenticated);
       if (!isAuthenticated) {
-        navigate(ROOT_ROUTES.LOGIN);
+        // Are we not authenticated because of a Google Auth session expiration?
+        const dueToGAuth = !!isSessionValid && !isGAuthSessionValid;
+
+        if (dueToGAuth) {
+          navigate(`${ROOT_ROUTES.LOGIN}?reason=gauth-session-expired`);
```
tyler-dane: Create a constant for the reason instead of just using the string here. This'll make it easier to test, access across other files, and rename it when things change

packages/backend/src/auth/controllers/auth.controller.ts
tyler-dane: Why couldn't that pattern be established?

packages/web/src/auth/gauth.util.ts
```diff
@@ -0,0 +1,19 @@
...
+export const validateGoogleAccessToken = async () => {
+  try {
+    const res = await fetch(`${ENV_WEB.API_BASEURL}/auth/google`, {
+      method: "GET",
+      credentials: "include",
+    });
+
+    if (!res.ok) return false;
+
+    const body = (await res.json()) as { valid: boolean };
```
tyler-dane: use the shared response type when casting here

packages/web/src/views/Login/Login.tsx
```diff
@@ -29,14 +43,23 @@ export const LoginView = () => {
...
+      toast.error("Google session expired, please login again");
+      clearSession();
+    } else {
+      checkSession().catch((e) => {
+        alert(e);
+        console.log(e);
+        clearSession();
+      });
+    }
+  }, [searchParams]);
```
tyler-dane: Do we not need to await `clearSession`?

packages/backend/src/auth/services/google.auth.service.ts
```diff
@@ -82,6 +117,21 @@ class GoogleAuthService {
...
+    const { token } = await this.oauthClient.getAccessToken();
+
+    if (!token) {
+      throw new BaseError(
+        "No access token",
+        "oauth client is missing access token. Probably needs a new refresh token to obtain a new access token",
+        Status.BAD_REQUEST,
+        false
+      );
+    }
```
tyler-dane, line 132: Rather than throwing a new `BaseError`, create an error in `error.constants` for this scenario.

Then throw it using the `error()` function like so:

```
throw error(AuthError.YourNewAuthError, "Access token not retrieved");
```

The `description` property in the error provides more info about the error. 
The second arg in the `error()` is meant to explain the result of the error not happening, which the client can use to determine next steps.

Try not to include too many implementation details in the result string, like "Probably needs a new refresh token to obtain a new access token." We don't want to send unnecessary info to the client. Instead, you can include that info in a debug log. 

 This'll make it easier to test, typecheck, and prevent bugs

packages/web/src/views/Login/Login.tsx
```diff
@@ -29,14 +43,23 @@ export const LoginView = () => {
...
-    checkSession().catch((e) => {
-      alert(e);
-      console.log(e);
-    });
-  }, []);
+    const reason = searchParams.get("reason");
+
+    if (reason === "gauth-session-expired") {
+      toast.error("Google session expired, please login again");
+      clearSession();
```
tyler-dane: 1. Will this toast even render for long, if we're immediately going to clear their session and cause a redirect? Do we need to wait for it to render before running `clearSession`? Is that better UX than just redirecting them immediately?

2. If you want to keep the toast, don't mention "Google session expired." Users don't care what a session is or that it's the Google connection that's requiring us to make them reauthenticate. Instead, say something like "It's been a while, so we're redirecting you back to the login page. This helps us keep your account safe"

packages/backend/src/auth/controllers/auth.controller.ts
```diff
@@ -63,6 +66,33 @@ class AuthController {
...
```
tyler-dane, line 95: 1. change `valid` to `isValid` for consistency with other booleans in the app.

2. Don't return `null`. Keep `isValid` as a boolean that always either returns true or false

packages/backend/src/auth/controllers/auth.controller.ts
that-one-arab: I remember it was mostly due to my confusion, trying to figure out how and under what circumstances its called in other route handlers, maybe if I spent more time I could have established it, I'll give it another go in this round of refactors.

that-one-arab: @tyler-dane So I checked the code. A google error is an error that's an instance of a `GaxiosError`. After looking through the function calls and examining google oauth library methods that are being called I did not notice any of them potentially throwing a `GaxiosError`

Regardless I still handled the error as a safety measure/sanity check. Please check [this commit](https://github.com/SwitchbackTech/compass/pull/212/commits/0f6b893bd8e1c2c17923b11ee98651747d130afb) and let me know what you think and if I should make any adjustments

tyler-dane: Glad you're thinking about edge cases and safety measures. But I'm afraid that if we don't know whether something will throw an error, we need to do more testing in order to be sure. If we know that no error will be thrown, then we shouldn't add error handling for it, which would just add to confusion for other devs.

I remove the fallback google error handling so we just know what we need. But if you investigate further and feel that it could occur, feel free to reintroduce it.

packages/backend/src/auth/controllers/auth.controller.ts
```diff
@@ -63,6 +66,36 @@ class AuthController {
...
+      // Upon receiving an access token, we know the session is valid
+      const accessToken = await gAuthClient.getAccessToken();
+
+      res.promise({ isValid: true });
+    } catch (error) {
+      const isBaseError = error instanceof BaseError;
+
+      if (isBaseError && error.result === "No access token") {
+        res.promise({ isValid: false, error: "Invalid Google Token" });
+        return;
```
tyler-dane: This conditional is why I prefer defining the error more explicitly. If we change the string in `getAccessToken` to be `Invalid access token`, then this catch won't work. If we had a predefined error object, we'd only need to change it in one place. This will be even more useful once we add integration tests, which could import our predefined errors.

packages/web/src/auth/gauth.util.ts
that-one-arab: @tyler-dane I'm not sure what type you're referring to, I tried searching for it inside `*.types` files but couldn't find something that looked like what you were specifying.

I opted to instead extract the type as a dedicated interface for now in 6e1a02f326da84680011a7a66b7d5157ec4e759f, please give me more info on what type you're referring to

packages/web/src/views/Login/Login.tsx
that-one-arab: @tyler-dane I opted not to await `clearSession` since its a void function and is always the last running code (meaning we always get the same last result) so from my knowledge no need to await it, also adding in async wrapper would be a little ugly.

I wanted to just prepend the statement with `void` keyword (`void clearSession();`) but after rethinking, I think we should indeed use await to avoid creating traps for future devs when working on this part of the code.

Let me know if you prefer this approach in edf1a815ce355d6f565bec2f47c72d33e236b9f6 or if we should just void the statement

packages/web/src/views/Login/Login.tsx
that-one-arab: 1. I'm not sure what you're referring to here about the redirect part or the toast rendering duration. From my examination, toast rendering duration is fine (default duration for `react-toastify`) and no redirecting occurs after clearing their session. Please clarify this part for me
2. Agreed. I used your example in 618aea787db0f7ac9c84d1fbe4304e29383b4bd7

that-one-arab:
@tyler-dane Done a round of refactors, responded to some comments that need your input and/or clarifications. Please let me know what you think

packages/web/src/views/Login/Login.tsx
tyler-dane: OK, that makes sense. Good with either, so I'm just going to resolve this

packages/web/src/views/Login/Login.tsx
tyler-dane: You're right, got this one mixed up with `ProtectedRoutes.tsx`

packages/web/src/auth/gauth.util.ts
tyler-dane: I meant to say that we should create a type for the backend rAPI response (stored in @core) and then use that type here, so that the frontend is in-sync with the backend. I've implemented that in a recent commit, so resolving

tyler-dane:
> @tyler-dane Done a round of refactors, responded to some comments that need your input and/or clarifications. Please let me know what you think

Looks good, just did a few refactors. There's currently a bug where a user gets into an infinite loop if they're on the login page and then tries to go to the root route before finishing the auth flow.

After a little more QA I will either fix that as part of this PR or just merge this and fix it in a separate bug.

tyler-dane:
Finished with cleanup, QA, and updating PR notes above.

I tweaked a few of your initial changes for practical reasons:

I removed the toast. This was working in isolation. But due to our duplicate renders, multiple toasts were appearing. I decided to simply remove them instead of increasing the scope of this PR so that we can focus on higher priority issues.

I used `axios` and our custom Auth API instead of `fetch` to reduce the confusion that'll come with having multiple requests implementations

PR #215, opened 2025-01-08
🐛 Fix: Consistently add `user` field to GridEvent

that-one-arab:
This PR is a followup to [this PR](https://github.com/SwitchbackTech/compass/pull/214), which address bug #207 

It now adds the `user` field to the `event` object before being submitted to the API to ensure that the `user` field consistently exists inside the object.

It also updates how events are parsed and validated to improve type checking


packages/core/src/types/event.types.ts
```diff
@@ -55,7 +55,7 @@ export interface Schema_Event {
   startDate?: string;
   title?: string;
   updatedAt?: Date;
-  user?: string;
+  user?: string; // TODO: `user` should never be optional. An event should always be associated with a user.
 }
```
that-one-arab: We should reconsider the "optionality" of all the fields inside `Schema_Event`. Some fields do not make sense to be optionally provided. If all fields are optional, we significantly weaken the advantages of the strong typing that TS provides.

packages/web/src/views/Calendar/hooks/draft/useDraftUtil.ts
```diff
@@ -96,29 +98,33 @@ export const useDraftUtil = (
...
-          reduxDraftType,
-          reduxDraft?.startDate,
-          reduxDraft?.endDate
-        );
-        setDraft({ ...defaultDraft, isOpen: true });
-        return;
-      }
+    if (!isDrafting) {
+      return;
+    }
```
that-one-arab: A small refactor. Personally I find early return statements makes the code easier to read, but it can be reversed if you think otherwise

tyler-dane (changes requested):
### Summary
These changes address the immediate issue of the user not being included in the backend request. They do that by accessing the `userId` from the `useUser` context hook and then passing that `userId` throughout the different functions as an argument.

### Problems
This seems innocuous now, but I'm afraid if we continued this pattern for the other properties, we'd make it prohibitively difficult to maintain and test the different parts of the frontend. The UX performance would also degrade at some point. Here's why

**Difficult to maintain because:**
- The `user` event is passed around `useShortcuts`, `useDraftUtil`, `useSidebarUtil`, `MainGrid`, `AllDayRow`, and `event.util`....but it never changes in any of those places. Others will wonder, "Why pass an arg to a function if it's not used?" If this continues, it'll be very difficult to track what is needed in any place and what isn't.

**Degraded performance**:
The more properties we include, the more we have to include whenever dragging, dropping, or resizing -- all of which are very expensive operations. We also will have to be careful not to break memoization every time we include a new property, as it introduces the possibility of re-rendering the component unnecessarily.

**Difficult to test**:
By using the `useUser` hook, we're making regular components dependent on Context and SuperTokens. This would require us to wire up a mock context provider and mock the SuperTokens library. That's acceptable when we're testing authentication and session management. But the implementation of the event functions that were changed here shouldn't depend on the authentication or session management implementation.


### Suggested Improvements
- Instead of passing around the user arg, keep the event payload minimal during editing. Then add before it's included in the request to the backend. 
- Create a new type or extend the existing `Schema_Event` so that `user` is a required value. 
- In the backend, use zod to parse the payload using that type before saving in the DB. This'll ensure it never gets persisted without a user.

### Mental Model
Here is how I think about what pieces of the app are responsible for what

| Piece      | Responsibility | How | Where |
| ----------- | ----------- | ------ | ------ |
| draft event      | Quick, understandable, testable       | minimal schema | draft functions |
| event (frontend) |   exhaustive copy of event from backend     | normalizes GET response, exhaustive schema | Redux store
| middleware | convert draft -> regular event | adds/derives values & validates against types/schemas | frontend typescript functions
| event (backend) | validates event and persists | zod, node, mongoDB, exhaustive schema | /backend package

Visual summary of the above:

Frontend Data Flow

that-one-arab:
@tyler-dane  Thank you for your detailed response. I agree with the points you mentioned, your provided mental model image also helped a lot in solidifying them for me to digest. I'll make the necessary refactors and tag you once done.

that-one-arab:
@tyler-dane PR is ready for re-review. I addressed the points you gave, but I had conflicting thoughts about [d8564c6110e7787fe031b0e3607c973aa1b8d9c5](https://github.com/SwitchbackTech/compass/pull/215/commits/d8564c6110e7787fe031b0e3607c973aa1b8d9c5), let me know your thoughts about it and if you think the changes are suitable for the codebase

tyler-dane:
@that-one-arab This is looking better, thanks for making those improvements. 

There's a way to make sure both the backend and frontend are validating and using the same type/schema without making the code too complicated. I started implementing it and will push up either a POC or the finished version soon. 

For now, feel free to continue on the other issue you're working on

tyler-dane:
Going to do some more thorough QA before merging, but my recent commits demonstrate how we can use zod to make our event objects more type-safe without adding too much technical overhead.

that-one-arab:
@tyler-dane Sweet, great work man. I like the type inference from the Zod event validator, this will make maintenance easier

Issue #210, opened 2025-01-01
Event times are not accurate based on mouse position


tyler-dane:
### Where did this happen?

Local & Prod

### Expected Behavior

Given that I'm DNDing or resizing an event on the grid,
Then the event times should accurately reflect the event's position and size

### Current Behavior

When DNDing or resizing, the event rectangle and times do not always match the mouse's position, especially when the mouse is close to the beginning or end of the event time.

Notice here that the times don't adjust as the mouse moves, and how the position doesn't move after the times change.

https://github.com/user-attachments/assets/01e51752-60f8-4638-9bef-391ece6c956a

### Steps to Reproduce

1. Drag an event over the grid
2. While still holding the mouse down, move the mouse upwards slowly
3. Notice: the event's times/size aren't immediately updating correctly
4. Keep moving the mouse upwards
5. Notice: the event's size/times eventually update to the correct position

### Possible Solution

We're currently relying on React's rendering cycle to take the mouse's position and compute the event rectangle. This passive approach doesn't handle edge cases when the times change, but not enough to adjust the event's size/position in turn.

Instead, improve the snaping implementation, which flips the approach. That is, instead of waiting for React to decide when to change an event's position/size, it prevents a change until one is warranted. Then it tells the event how to render more actively. 

Reference how we're currently snapping here:
- `GridEventPreview.styled.ts:getItemStyles()`

Other resources to help with snapping:
- `dnd-grid`'s [`snapToGrid.js`](https://github.com/jamesdconklin/dnd-grid/blob/164d094088bbf8752a9de544e7812cb1ae32c092/src/snapToGrid.js#L26)
- `dnd-grid`'s [repo](https://github.com/jamesdconklin/dnd-grid), which also shows how to use grid constraints, render arbitrary child components, and reordering children
- `dnd-grid`'s [live demo](https://jamesdconklin.github.io/dnd-grid/)
- "[How can I snap a game object's position to a grid?](https://gamedev.stackexchange.com/questions/33140/how-can-i-snap-a-game-objects-position-to-a-grid)"

### Context

Dragging, dropping, and resizing are some of the most used actions for a calendar. Not having a good UX for this makes uses not want to use the app

that-one-arab:
Assign me this please

PR #218, opened 2025-01-16
🐛 Fix: Implement better snapping when dragging events from sidebar to grid

that-one-arab:
## Description

Fixes [issue](https://github.com/SwitchbackTech/compass/issues/210)

This PR handles implementing more accurate snapping when dragging someday events to the grid.

The issue the PR addresses happens due to a calculation mismatch between 2 methods: `GridEventPreview:getTimePreview()` and `GridEventPreview/styled:getItemStyles()`. This is due to the methods relying on different calculation variables, producing different results in certain mouse movement edge cases.

The PR fixes that by matching the calculation logic inside the `getItemStyles` method with the `getTimePreview` method (it does not do it directly, because as a small refactor, we extracted snapping logic from `getItemStyles` inside its dedicated `snapToGrid` function. See `snapToGrid`)

To achieve that, we needed to feed calculation logic variable dependencies inside the tree, hence, why we needed to prop drill.

This is a common pattern inside the application and it is anti best practice. Ideally we should be using something like React context to feed variables down the tree without prop drilling.

I thought this type of refactor is out of scope of this PR, and I also felt like I was spending too much time on this PR so I wanted to tie things up and consult you from here.

As a side quest I snuck in an extra feature, horizontal snapping! (I took inspiration from google calendar and felt like we should implement it here too)

## Results
Please take a look at the before and after, you will notice the following:
- Event time accuracy issue is resolved
- Horizontal (column based) snapping is implemented

#### Before
[vertical_snap_before.webm](https://github.com/user-attachments/assets/2e727453-6d59-4d84-8f46-681f0160d8dd)

#### After
[vertical_snap_after.webm](https://github.com/user-attachments/assets/3f93afe8-d593-4b4f-9daf-c716ef99b8dd)

### DIFF
--- a/packages/web/src/views/Calendar/components/Event/Grid/GridEventPreview/snap.grid.test.ts
+++ b/packages/web/src/views/Calendar/components/Event/Grid/GridEventPreview/snap.grid.test.ts
@@ -2,6 +2,7 @@ import {
   MeasureableElement,
   Measurements_Grid,
 } from "@web/views/Calendar/hooks/grid/useGridLayout";
+import { MouseCoords } from "@web/views/Calendar/hooks/draft/useMousePosition";
 
 import { snapToGrid } from "./snap.grid";
 describe("snapToGrid", () => {
@@ -28,9 +29,12 @@ describe("snapToGrid", () => {
   };
 
   it("returns the cursor position if measurements are not available", () => {
+    const mouseCoords: MouseCoords = { x: 200, y: 300 };
+    const scrollTop = 0;
+
     const result = snapToGrid(
-      200,
-      300,
+      mouseCoords.x,
+      mouseCoords.y,
       {
         mainGrid: null,
         allDayRow: null,
@@ -40,14 +44,24 @@ describe("snapToGrid", () => {
           console.log("foo", elem);
         },
       },
-      0
+      scrollTop
     );
-    expect(result).toEqual([200, 300]);
+
+    expect(result).toEqual([mouseCoords.x, mouseCoords.y]);
   });
 
   it("returns the cursor position if cursor is not within the grid", () => {
-    const result = snapToGrid(200, 50, measurements, 0);
-    expect(result).toEqual([200, 50]);
+    const mouseCoords: MouseCoords = { x: 200, y: 50 };
+    const scrollTop = 0;
+
+    const result = snapToGrid(
+      mouseCoords.x,
+      mouseCoords.y,
+      measurements,
+      scrollTop
+    );
+
+    expect(result).toEqual([mouseCoords.x, mouseCoords.y]);
   });
 
   it("snaps the cursor position to the nearest grid interval", () => {
--- a/packages/web/src/views/Calendar/components/Event/Grid/GridEventPreview/snap.grid.test.ts
+++ b/packages/web/src/views/Calendar/components/Event/Grid/GridEventPreview/snap.grid.test.ts
@@ -4,7 +4,9 @@ import {
 } from "@web/views/Calendar/hooks/grid/useGridLayout";
 import { MouseCoords } from "@web/views/Calendar/hooks/draft/useMousePosition";
 
-import { snapToGrid } from "./snap.grid";
+import { snapToGrid, TIME_COLUMN_WIDTH } from "./snap.grid";
+
+// 7 day main grid
 describe("snapToGrid. 7 day grid", () => {
   const measurements: Measurements_Grid = {
     mainGrid: {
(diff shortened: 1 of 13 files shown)


packages/web/src/views/Calendar/components/Event/Grid/GridEventPreview/snap.grid.test.ts
```diff
@@ -0,0 +1,81 @@
...
+        throw new Error("Function not implemented.");
+      },
+    },
+    hourHeight: 60,
+    colWidths: [100, 100, 100, 100, 100, 100, 100],
+    allDayRow: null,
+    remeasure: (elem: MeasureableElement) => {
+      console.log("remeasuring element", elem);
+    },
+  };
```
tyler-dane, line 31: Passing around `measurements` leads to a lot of setup for these tests, which makes it harder to follow. I'm OK continuing this pattern for now, but it'd be good to do a follow-up sometime to extract this into a (Context) Provider, so that the test setup is more minimal

packages/web/src/views/Calendar/components/Event/Grid/GridEventPreview/snap.grid.test.ts
```diff
@@ -0,0 +1,81 @@
+import {
+  MeasureableElement,
+  Measurements_Grid,
+} from "@web/views/Calendar/hooks/grid/useGridLayout";
+
+import { snapToGrid } from "./snap.grid";
```
tyler-dane: I recommend just testing `snapToGrid`, as that lets us effectively cover the X and Y coordinate logic without having to also test the `snapYToGrid` and `snapXToGrid` separately

packages/web/src/views/Calendar/components/Event/Grid/GridEventPreview/snap.grid.ts
```diff
@@ -1,7 +1,87 @@
...
+  measurements: Measurements_Grid
+): number => {
+  if (!measurements.mainGrid) return cursorX; // TS guard
+
+  // `mainGrid.left` includes half the width of time column. It does not start at the first day column.
+  // due to that, we need to account for its width, otherwise snappedX will be off.
+  // Ideally we should fix how `mainGrid.left` is calculated to not include half the width of time column and
+  // include only the grid interactivity area (i.e. day columns).
+  // For now, we estimate the width of the time column to be 55px
+  const TIME_COLUMN_WIDTH = 55;
```
tyler-dane: Agreed here. Something we should address after this PR is merged and we have more tests. No need to block PR for it now, though

tyler-dane (commented):
Nice job here! Snapping is tricky.

Functionality is good. Feels like a solid UX improvement.

I have a few requests.

### Testing
The first is to add unit tests for the snapping logic. I added a test file and stubbed out a few uses cases to get it started. Feel free to adjust those as needed.

The reason I want to include tests is so we can more easily extend this snapping pattern for other use-cases. We could improve snapping for the timed week event, the all day row, and between sections of the sidebar. With that in mind, it'll be important to write these tests in a way that are 

1. **Understandable**: It should be relatively easy for a newcomer to read the tests, understand their intent, make changes, and write similar tests
2. **Focused**: The view should be decoupled/abstracted from the test as much as possible. The tests should work whether we're snapping on a 7 day week, a 4 day week, or a day view. Minimal changes and setup are acceptable, but out tests and implementation shouldn't be tightly connected to our current view. 
3. **Unit tests**: Since our view is going to change, let's just focus on the logic. No need to write interaction or e2e tests for these. Those'll come in a separate effort.

### Refactoring
You're right about how we shouldn't prop drill the measurements, refs, etc as much. Feel free to refactor as you go while you add tests. But take care not to go too far down this route. We can always do a separate refactor PR once we have the tests.

that-one-arab:
@tyler-dane I pushed several commits that addresses implementing test cases for `snapToGrid`

Finding a balance between making the test cases readable and not leaking inner code implementation was tricky, hopefully I did a good job here, let me know your feedback on this.

I thought of implementing a pattern where each describe block represents a certain calendar view (7 day, 3 day). If we were to implement this pattern, we would have a dedicated describe block for 7 day view, and a dedicated describe block for 3 day view. Let me know your feedback on this pattern.

Once you provide feedback I'll make final cleanups and/or improvements

tyler-dane:
Nice job @that-one-arab, now we have unit tests that explain how the snapping should work and will make improving it a lot easier.

I made some small tweaks for readability. I think splitting the test up between horizontal and vertical snapping is the way to go. That way, we don't have to keep coupling the tests to a 7 day main grid like we hare now. As we add more views, we could just create an array of views (1 day, 4 day, 7 day), and loop over them, running each of the existing tests we have. 

Screenshot 2025-01-19 at 8 04 43 AM

Issue #140, opened 2024-10-01
🚧 Create Actions menu and render on right-click [WIP]


tyler-dane:
**Implementation**
- [ ] Don't trigger mouse down event when R-Clicking

PR #234, opened 2025-01-28
✨ Feat(web): handle overlapping events

that-one-arab:
This PR handles detecting and displaying overlapping events so that they're not on top of one another, effectively being hidden and causing bad UX when attempting to view or drag/edit them.

[handle_overlapping_events.webm](https://github.com/user-attachments/assets/5d20983c-eed9-44a2-b3ee-9ae5e4800a5e)

### DIFF
--- a/packages/web/src/common/utils/event.util.test.ts
+++ b/packages/web/src/common/utils/event.util.test.ts
@@ -1,4 +1,8 @@
-import { isEventInRange } from "./event.util";
+import {
+  isEventInRange,
+  adjustOverlappingEvents,
+  assembleGridEvent,
+} from "./event.util";
 
 describe("isEventInRange", () => {
   it("returns true if event is in range", () => {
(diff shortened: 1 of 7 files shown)


that-one-arab:
@tyler-dane This PR is ready to be merged, but I prefer implementing unit tests before merging, seeking your guidance on how to proceed here.

packages/web/src/common/utils/event.util.ts
```diff
@@ -90,28 +90,38 @@ export const getDefaultEvent = (
   startDate?: string,
   endDate?: string
 ): Schema_GridEvent | null => {
+  const defaultEvent: Schema_GridEvent = {
+    priority: Priorities.UNASSIGNED,
+    position: {
+      isOverlapping: false,
```
that-one-arab, line 121: `isOverlapping` may appear redundant, but it adds significant boost to code readability in cases like `packages/web/src/common/utils/grid.util.ts:441`

packages/web/src/common/utils/event.util.ts
```diff
@@ -90,28 +90,38 @@ export const getDefaultEvent = (
   startDate?: string,
   endDate?: string
 ): Schema_GridEvent | null => {
+  const defaultEvent: Schema_GridEvent = {
+    priority: Priorities.UNASSIGNED,
+    position: {
```
that-one-arab, line 120: I made position to be a required object on `Schema_GridEvent` because I feel like we should start making a lot of these properties inside the interface to be required so we can make developer expectations simpler rather than being forced to conditionally check the existence of a property before dealing with it (EG: `(if (position && position.isOverlapping) { // Do Stuff } `

packages/web/src/ducks/events/selectors/event.selectors.ts
```diff
@@ -34,9 +34,16 @@ export const selectGridEvents = createSelector(
...
+    const weekEvents: Schema_GridEvent[] = weekEventsMapped
+      .filter((e: Schema_Event) => e !== undefined && !e.isAllDay)
+      .map((e) => ({
+        ...e,
+        position: {
+          isOverlapping: false,
+          widthMultiplier: 1,
+          horizontalOrder: 0,
+        },
+      }));
```
that-one-arab: Here we need to set the `position` object to default values (to comply with the interface)

I believe setting it here when fetching the events from the redux store is appropriate, considering that the redux store itself does not need the `position` object, but we start to need it once we start using it in the UI, right after fetching it from the store.

that-one-arab: Upon reading, I believe we can make a good refactor to make default position object values as a constant variable

tyler-dane:
Upon first load, this is what I'm seeing. 

isOverlapping

After changing  `event.position.isOverlapping && ...` to `event.position?.isOverlapping && ...` in `grid.util.ts:getAbsoluteLeftPosition` it went away. 

I saw that you made `event.position` required and I wasn't getting a type error on that line, but something must still be wrong.

We can't merge it with this issue - can you investigate?

tyler-dane:
Another blocking issue: The form position isn't working for grid events. They're always in the top left now.
form

tyler-dane:
> @tyler-dane This PR is ready to be merged, but I prefer implementing unit tests before merging, seeking your guidance on how to proceed here.

Agree, let's add some testing for this new algorithm. 

I'd focus on testing `adjustOverlappingEvents`, since that works independent of our store and selector implementation. Add tests to validate:
- sorting by time
- sorting by title as fallback

Don't try to validate `isOverlapping` or `widthMultiplier`. We only care that the `horizontal` order is correct for any given array of events.

packages/web/src/ducks/events/selectors/event.selectors.ts
tyler-dane: Agree, this is a good spot to add these overlapping properties. 

Feel free to refactor to make it more readable as you see fit

packages/web/src/common/utils/event.util.ts
tyler-dane: Agreed, thanks for making this more type safe

that-one-arab:
> Another blocking issue: The form position isn't working for grid events. They're always in the top left now. form

I was unable to reproduce both issues reported by you, could you please share reproduction steps with me?

tyler-dane:
> > Another blocking issue: The form position isn't working for grid events. They're always in the top left now. form
> 
> I was unable to reproduce both issues reported by you, could you please share reproduction steps with me?

I just checked out to this PR in the CLI and tried loading the app in dev. 

Try logging out or creating more events from `main` and then coming back to this PR's branch to reproduce.


https://github.com/user-attachments/assets/2c5351f1-2cfc-4e54-b4f7-7963e39b29e5

https://github.com/user-attachments/assets/b77ab03c-a37e-469a-b5eb-03e0be288b20

that-one-arab:
@tyler-dane Still couldn't reproduce it. I noticed that my branch was behind in some changes, I thought maybe this was the reason, but even after pulling and rebasing I still could not reproduce
[bug_reproduce_attempt.webm](https://github.com/user-attachments/assets/8d23194f-e21c-4097-bfa7-bb88ca337049)

Maybe we can set up a meeting to try and reproduce it live?

that-one-arab:
@tyler-dane I did some forced updating and finalized my refactors and unit tests. Please let me know if you are still reproducing this, if yes, maybe we can schedule a meeting to see this bug in action

packages/web/src/common/utils/event.util.test.ts
```diff
@@ -18,3 +22,198 @@ describe("isEventInRange", () => {
     expect(isEventInRange(event, dates)).toBe(false);
   });
 });
+
+describe("adjustOverlappingEvents", () => {
+  // Helper function that constructs a date string from hour:minute input.
+  // Makes date time strings easier to read
+  const time = (time: string) => `2025-01-27T${time}:00+03:00`;
```
tyler-dane, line 29: nice

tyler-dane:
> @tyler-dane I did some forced updating and finalized my refactors and unit tests. Please let me know if you are still reproducing this, if yes, maybe we can schedule a meeting to see this bug in action

I have a feeling it's related to how you're testing. Perhaps you're using a local account without any pre-existing calendar events? 

Regardless, I want to avoid scheduling a debugging meeting due to my restrictive schedule. I pushed up a commit to make that check optional, which is 'fixing' it for me. I'd like to get this PR in so you can start on #140. Going to merge this and do a little more debugging on my end; might do a follow-up cleanup PR if I find anything that'd help the error I was seeing. 

Overall nice job on the functionality, tests, and timeliness!

Issue #229, opened 2025-01-25
Change cursor when dragging grid event


tyler-dane:
## Current Behavior
When clicking and dragging an event on the grid the cursor changes from `default` to `grab`.

https://github.com/user-attachments/assets/7621237b-758e-46a6-ab59-8d12994bd005

Notice how the cursor changes when it gets close to the event borders.

## Expected Behavior
The cursor behavior mimics Google Calendar's implementation

https://github.com/user-attachments/assets/27ef842a-f46c-44ed-9fb8-397c918e25e0

## Acceptance Criteria
- [ ] When hovering over an event, the cursor changes from `default` to `pointer.`
- [ ] When mousing down an event and immediately dragging it, the cursor changes from `pointer` to `move`
- [ ] When clicking an event and holding the cursor in the same position for ~1 second, the cursor changes from `pointer` to `move.` (Doesn't have to be _exactly_ 1 second - experiment with what feels right.)
- [ ] Works for existing grid events
- [ ] Works for draft grid events (when the event hasn't been saved to the DB yet)

## Use-Case
This will improve the dragging UX, a very common task for users. By having a consistent drag experience that subtly suggests what to do next, they will reschedule their event quicker and will be more likely to use the app.

## Resources
https://developer.mozilla.org/en-US/docs/Web/CSS/cursor

MrPand-21:
I can do that in a week,

Are we trying to change the cursor from grabbing icon to that of google calendar or preventing cursor to change.

tyler-dane:
Hey @MrPand-21, thanks for volunteering and providing a time estimate for when it'll be done!  I set the end date to be February 2 so we can track that in the project. 

I also added a few more details to the issue. LMK if it's still unclear.

MrPand-21:
I am done with the issue yet I am not sure how to send a pr, git push on a new branch alerts "Permission to SwitchbackTech/compass.git denied"

tyler-dane:
@MrPand-21 I assigned this to myself for a few reasons
1) This is a timely issue that needs to get done this quarter, and we've gone well over the initial time budget
2) The outstanding issues I mentioned in the comments of #239 
3) The solution to this issue impacts what I'm working on in #199 

I appreciate you being willing to help. I'd be happy to work with you on an issue for Q2, which I hope will allow us to be more flexible with the timing as you get more familiar with the codebase.

PR #239, opened 2025-01-30
styled draft & event cursor on movements

MrPand-21:


packages/web/src/views/Calendar/components/Event/Grid/GridEvent/GridEvent.tsx
```diff
@@ -58,6 +61,10 @@ const _GridEvent = (
   const { component } = weekProps;
 
   const isInPast = dayjs().isAfter(dayjs(_event.endDate));
+  const [isDragPaused, setIsDragPaused] = useState(false);
+  const dragAndPauseThreshold = 480; // in ms, change accordingly to responsiveness
+  const timer = useRef<NodeJS.Timeout | null>(null);
```
tyler-dane, line 66: Typing it with the NodeJS library is misleading, because it suggests the code is running in a Node.js environment, rather than as a web app.

tyler-dane (commented):
This is a good start. 

It has the following issues:
1. The cursor changes back from `move` to `pointer` if it is not moved. It should stay `move` as long as the mouse is down.

2. The cursor changes from `move` to `default` if the mouse extends the event's rectangle sides. 

Issues shown here:
https://github.com/user-attachments/assets/703ffc60-e667-470a-badf-94ddf5cc7d03


To address, I'd consider making your changes outside of the `GridEvent`, higher up in the tree. `useGridMouseMove` might be a good place to start.

packages/web/src/views/Calendar/components/Event/Grid/GridEvent/GridEvent.tsx
MrPand-21: I am trying to understand the whole architecture and therefore it would take some time probably to spot the right places but i will try, also how does the requirement

> When clicking an event and holding the cursor in the same position for ~1 second, the cursor changes from pointer to move. (Doesn't have to be exactly 1 second - experiment with what feels right.)

different from 

> When mousing down an event and immediately dragging it, the cursor changes from pointer to move

If user drags the event, cursor will be move all the time, then what is the use of the first requirement? I thought if the user drags and pauses cursor would switch back to pointer. In the case user didn't clicked the event, chaning cursor to move would be misleading, so this is not the requisite either. Then what is it? 

Thanks,

tyler-dane: Sure, it definitely takes some time to get familiar so no worries.

I'll try to clarify the requirements a little better.

Basically, it should work like how it works in Google Calendar.

If a user mouses down on an event and does not move the mouse, then the cursor should stay `pointer` for  some time delay and then change to `move`
- why: the delay gives them a chance to mouse up and trigger the event form to open. The change to `move` gives them feedback that they're now dragging an event, which encourages them to move the mouse. It also sets the expectation that the form will not open after they mouse up.

If a user mouses down on an event and immediately moves the mouse, then the cursor should change to `move`
- why: to indicate that they're moving the event, and shows that they don't always have to wait for the delay to finish before doing so

tyler-dane:
Closing based on comments in #229

Issue #236, opened 2025-01-28
Improve overlapping events: spacing and drag-and-drop UX


tyler-dane:
Follow-up from #234 (#213). That PR focused on the baseline functionality of overlapping events. This one focuses on providing better styling and UX for these overlapping events

- [x] Add space between overlapping events. Reference how Outlook does this.
  - This will make it easier to see and click events that overlap
- [x] When dragging an overlapping event, the event width expands to the full column.
  - Without this, the dragged event's width doesn't match the cursor or column lines, which makes repositioning it feel awkward.
  - Reference: Outlook, Gcal, Notion Cal, Vimcal

tyler-dane:
@that-one-arab I asked you to work on #140 , but I think it makes sense to wrap this one up beforehand. That work?

that-one-arab:
@tyler-dane Yes that's better, on it.

PR #240, opened 2025-01-31
✨ Feat: Better UX for overlapping events

that-one-arab:
Addresses [issue](https://github.com/SwitchbackTech/compass/issues/236)

[overlapping_events_better_ux.webm](https://github.com/user-attachments/assets/2fbec91a-9f84-41a0-82d7-571491b636e5)


packages/web/src/common/utils/grid.util.ts
```diff
@@ -407,16 +409,32 @@ export const getLeftPosition = (
...
-    eventWidth
+    eventWidth,
+    isDraft
   );
 
+  if (isDraft || !event?.isAllDay) {
+    left += GRID_MARGIN_LEFT;
+  }
+
+  if (event?.position.isOverlapping) {
```
tyler-dane: This line was causing the same error as before when trying locally. Please add all-day events to your testing account. This will help us catch bugs earlier

that-one-arab: Ah my bad thanks for the headsup

Issue #241, opened 2025-01-31
Scroll up/down when dragging grid event


tyler-dane:
### Feature Title

Smart Scroll During Drag

### Feature Description

The dragging a timed event up or down outside of the current view, the draft is no longer seen. Instead, we should automatically scroll so that the draft is always visible

### Use Case

This'll make users more comfortable dragging events, because they'll always be able to see them

### Additional Context

This will likely require some logic to identify if the current mouse position is in the current view port. If not, scroll upwards/downwards until it is

tyler-dane:
Hey @that-one-arab , heads up that that I'm working on refactoring how drafts are created and updated in #255, which may result in conflicts with your implementation here. I'm trying to keep the changes contained to minimize that scenario. 

Normally I'd encourage you to refactor surrounding code as you do your implementations in an effort to always leave the code more clean and maintainable, but since our cleanups might differ, in this case you don't need to do that. Instead, feel free to share suggestions on how to change related code in the comments of #255

that-one-arab:
@tyler-dane Hey, I am working on this issue and I think I reached a point where I am probably blocked from working on it.

In [262](https://github.com/SwitchbackTech/compass/pull/262) `useDragEventSmartScroll:16` I have `if (!draftEvent) return;`

This will be corrected to something like `if (draftEvent.status !== 'dragging') return;` but regardless, I need to know the type of action being performed on the draft event (resizing, dragging, or simply editing)

Right now this isn't possible to do using redux selector, because in `MainGridEvents:49` we always dispatch `draftSlice.actions.startDragging` regardless of whether we are actually dragging the event, which sets the event status to `dragging`.

So I would either need to do something really ugly (like passing `useDraftUtil:draftState` around to get the state from this object, or somehow refactor the current implementation to allow me to get the draft event status from redux store.

But since you already warned me that you are doing large refactors with how draft events work (I also noticed your draft PR here https://github.com/SwitchbackTech/compass/pull/255/ I am unsure on how to proceed

Please advice.

tyler-dane:
Hey @that-one-arab , thanks for the heads up about being blocked. 

It sounds like the core issue is that in order to implement this feature correctly, you need to be able to distinguish when the user is actively dragging an event (has his mouse down on an event) and when they are simply editing an event (resizing or looking at an open form).

#255 just got merged, which I think will make working with this easier. However, the core issue will still be there.

I share your hesitancy to do something really ugly.

[...]

PR #247, opened 2025-02-04
✨ Feat: Custom actions context menu

that-one-arab:
This PR handles implementing a custom context menu that triggers when users right click an event, as well as general code refactors.

It supports for all time events and timed events (no support for someday events yet)

It contains:
- Priority edit buttons: Allows you to quickly edit an event's priority
- Edit button: triggers opening the draft event form
- Delete button: deletes the event

image

[custom_context_menu.webm](https://github.com/user-attachments/assets/0cddbcf2-662d-4217-908c-b747406a6aa2)


tyler-dane:
Hey @that-one-arab , I see you're finding solutions to make this work. But I'm concerned that we're reinventing the wheel and adding unnecessary complexity along the way. Specifically, calculating the x,y coordiantes, creating context menu, and adding an invisible backdrop. 

Please check out [floating-ui](https://floating-ui.com/), which can solve a lot of these problems out-of-the-box and in a more reusable fashion. We're already using this for the forms, and could use it for other elements like tooltips (for shortcuts) and select menus (for time inputs).

Although what you have so far might work, I'd rather go with the more resusable solution that requires less code and maintenance.

that-one-arab:
> Hey @that-one-arab , I see you're finding solutions to make this work. But I'm concerned that we're reinventing the wheel and adding unnecessary complexity along the way. Specifically, calculating the x,y coordiantes, creating context menu, and adding an invisible backdrop.
> 
> Please check out [floating-ui](https://floating-ui.com/), which can solve a lot of these problems out-of-the-box and in a more reusable fashion. We're already using this for the forms, and could use it for other elements like tooltips (for shortcuts) and select menus (for time inputs).
> 
> Although what you have so far might work, I'd rather go with the more resusable solution that requires less code and maintenance.
> 
> If we can use a good library like floating UI to allow us to just focus on the unique aspects of a context menu in Compass (its styling and elements), that'd be a big win

Agreed. I wasn't aware of the library, I am checking it out and I like the reliability of it and the complex positioning scenarios it handles out of the box. I will refactor the PR accordingly.

packages/web/src/views/Calendar/components/Grid/MainGrid/MainGridEvents.tsx
```diff
@@ -47,6 +47,8 @@ export const MainGridEvents = ({ measurements, weekProps }: Props) => {
 
   const editTimedEvent = (event: Schema_GridEvent) => {
     dispatch(
+      // TODO: `actions.startDragging` is confusing, it appears to mix concerns for both dragging
```
tyler-dane: Agreed. Some of my changes from my current draft PR will improve how this works

tyler-dane: PS This comment is just be a harmless note-to-self in a draft PR, but it's a good opportunity to share my thoughts on comments in general: https://docs.compasscalendar.com/docs/contribute/convention-guide#comments.

Go ahead and keep using TODO comments during your WIP PRs if it helps your workflow, but please remove them before submitting the PR. If they're still relevant at the time of publishing your PR, you could add an issue or inline PR comment like I'm doing here so we don't miss out on your thoughts

that-one-arab: @tyler-dane Nice catch yes I wasn't aware about the comments policy, and I happen to agree to your philosophy regarding it. I'll make sure to clean it up.

packages/web/src/views/Calendar/components/Grid/contextmenu/InvisibleBackdrop.tsx
tyler-dane, line 1: What is the purpose of this backdrop? If it's to register outside clicks and close the menu, consider floating-ui's `onOpenChange` instead, which takes care of that for you.

that-one-arab: yes good catch, I meant to clean this up.

packages/web/src/views/Forms/SomedayEventForm/SomedayEventForm.tsx
```diff
@@ -116,7 +117,9 @@ export const SomedayEventForm: React.FC<FormProps> = ({
       role="form"
     >
       <StyledIconRow>
-        <DeleteIcon onDelete={onDelete} title="Delete Someday Event" />
+        <IconButton onClick={onDelete} aria-label="Delete Someday Event">
+          <Trash />
+        </IconButton>
```
that-one-arab, line 122: In the future we should refactor other components that use the same pattern the `Delete` icon does.

that-one-arab:
@tyler-dane PR is ready! Please let me know your feedback.

tyler-dane:
Thanks for getting this to a stopping point after we chatted descoping someday events yesterday. 
Added some cleanup commits and explained reasoning in the commit descriptions. Functionality looks good! 

Noticed this introduces bug #256, but that's not a show-stopper. Please move on to #241.

Issue #199, opened 2024-12-24
Creating a new sidebar event causes the order to change


tyler-dane:
### Where did this happen?

Production app, Local code with an up-to-date main branch

### Expected Behavior

When creating a new sidebar event, the other sidebar events should have mostly the same position (a few might need to change to make room for the new one)

### Current Behavior

When creating a new sidebar event, the order of the events changes in unpredictable ways

### Steps to Reproduce

1. Create 5 sidebar events
2. Create another new sidebar event
3. Notice: how the other 5 shifted position after the new one was created

### Context

This disorients users and makes them less likely to create sidebar events

that-one-arab:
@tyler-dane I could not reproduce this, can I get more clarification on this please? A video recording would be nice too

![Image](https://github.com/user-attachments/assets/e45c833d-dbe5-4a0d-99e4-e34eb4ff64ab)

tyler-dane:
Also had trouble reproducing this. Perhaps it was fixed during the refactor of #275 

Closing, will re-open if I can reproduce

PR #253, opened 2025-02-06
⚠️ Feat(POC): Trigger event actions on intuitive mouse events

tyler-dane:
This is a proof-of-concept of how to trigger events using global event listeners and state, with a focus on mouse events. It provides a UX that is more aligned with user expectations. It also enables future enhancements and fixes, including #199. 

## Summary
Further separates concerns between mouse, clicks, events, and drafts. 

### 1. Delegates form display to the `floating-ui` library
Whenever a user clicks on a grid, there are many potential next steps that might occur: creating a new event, closing the previous event form, dragging, or resizing. We previously had to add conditionals into child components (like the header and sidebar) that would check the state of the app so help us identify which scenario we were in at the time of the click. 

Now we're using `floating-ui`'s `onOpenChange` callback and `useDismiss` hook to do that for us. We configured the forms to close whenever the user clicks `ESC` or clicks outside of the form. This allows us to stop polluting our components with form-checking behavior.

### 2.  Decouples the form and draft.
We were previously tracking whether an event should open its form by adding an `isOpen` boolean to every grid draft. The downside of that approach is that although only one form can be open at a time, we had to add that property to every event. Thanks to the change above, we could now accomplish this using one state, `isEventForm`, that is separate from individual events. This makes event state more minimal and also simplifies event creation and teardown.

### 3. Changes mouse listeners and events
This is how mouse handlers previously worked from a user's perspective:
| Action    | Result |
| -------- | ------- |
| mouse down | created new event and immediately started resizing the end date    |
| mouse down on grid when drafting a different event  | draft discarded    |
| mouseup | form opened     |
| click    | n/a    |

This is how they work now:
| Action    | Result |
| -------- | ------- |
| mouse down | nothing    |
| move while mouse is down | start resizing |
| mouse down on grid when drafting a different event  | n/a    |
| mouseup after mousing down and moving | resizing stops, form opens     |
| click on grid | new draft created and form opens | 
| click on grid when drafting a different event   | draft discarded    |


tyler-dane:
closing in favor of #255

PR #268, opened 2025-02-21
✨ Feat: Update event cursor css when dragging

that-one-arab:
## Description
Closes [issue](https://github.com/SwitchbackTech/compass/issues/229)

When clicking and dragging an event on the grid the cursor changes from default to grab.

This PR implements the right cursor icons for each respective click state

This will improve the dragging UX, a very common task for users. By having a consistent drag experience that subtly suggests what to do next, they will reschedule their event quicker and will be more likely to use the app.

The PR ticks the below criteria
- When hovering over an event, the cursor changes from `default` to `pointer.`
- When mousing down an event and immediately dragging it, the cursor changes from `pointer` to `move`
- When clicking an event and holding the cursor in the same position for 0.75 second, the cursor changes from `pointer` to `move.`
- Works for existing grid events
- Works for draft grid events (when the event hasn't been saved to the DB yet)

The PR also handles some general code refactors.

#### Before
[mouse_hold_delay_feat_before.webm](https://github.com/user-attachments/assets/c106831f-ae40-488c-9cc7-0dabae690451)

#### After
[mouse_hold_delay_feat.webm](https://github.com/user-attachments/assets/5e3272b5-4553-4f91-8377-f04cbf2340fc)


packages/web/src/views/Calendar/layout.constants.ts
```diff
@@ -34,3 +34,5 @@ export const GRID_X_PADDING_TOTAL =
 export const SIDEBAR_MONTH_HEIGHT = 275;
 export const SIDEBAR_OPEN_WIDTH = 350;
 export const SIDEBAR_X_START = SIDEBAR_OPEN_WIDTH + GRID_X_START;
+
+export const GRID_EVENT_MOUSE_HOLD_DELAY = 750;
```
tyler-dane: nitpick: `layout.constants` is meant for detailing the page's layout, which'll come in handy as we make the page more dynamic and responsive. The values are shared across many parts of the app, so have a dedicated file for it is helpful.

This value isn't related to the layout, and is only used in one place, so it can live there instead.

that-one-arab: @tyler-dane Not sure I follow, from what I understood you are saying that the `GRID_EVENT_MOUSE_HOLD_DELAY` doesn't belong inside this constants file, but I didn't understand where are you suggesting we place it.

Maybe creating a new `grid.constants.ts`?

tyler-dane: Yes, that `GRID_EVENT_MOUSE_HOLD_DELAY` doesn't relate the the page's layout, so it shouldn't be in `layout.constants`. Just updated it to move it inside the file that uses it. If there are other parts of the app that need that value, we move it to a different dedicated constants file

tyler-dane:
Nice work! 🚀

PR #301, opened 2025-03-12
♻️ Refactor: Extract sync code into separate microservice (NestJs)

### LINKED ISSUE #299: Display recurring events that were created in Google Calendar after initial import
While we're importing all recurring events during the initial import, we also need to import new recurring events after a user creates them in google calendar.

tyler-dane:
Closes #299 

- [x] Extract sync code into separate directories based on functionality (import, watch, unwatch, etc)
- [x] Refactor sync functions into a class to reduce complexity (eg minimizes how many times a gcal client is instantiated or passed around)
- [ ] Extract the sync functionality into its own service (initially this'll run in Node, but eventually Docker will be added as well)
- [ ] Install msw for integration testing
- [ ] Trigger requests to new sync service and ensure it is returning a response
- [ ] Write tests with msw to validate the above
- [ ] Update sync service so that it correctly updates database after receiving a sync notification from Google Calendar
- [ ] Remove TODOs and unnecessary comments

## Known issues
Duplicate types between `@core` and this `sync` package
- Had difficulty sharing imports and module aliases between these two
- Since we will convert all backend packages into a new monorepo system (likely Nx), this is acceptable for now


tyler-dane:
Abandoning in factor of #306 . I got too carried away by using NestJS in this PR, which resulted in a lot of unexpected config issues related to jest and the lerna/npm monorepo. We will need to hold off on big changes to our stack and start transitioning more gracefully rather than sneaking them into feature PRs on a whim like this. Lesson learned.

Issue #298, opened 2025-03-11
Cannot DnD an event when its form window is open


that-one-arab:
### Where did this happen?

Local code with an up-to-date main branch

### Expected Behavior

When I click an event, it would open the event form window, then if I try to drag and drop the event, the event window should close and I should be able to drag and drop it.

### Current Behavior

I am unable to drag and drop the event

### Steps to Reproduce

1. Create an event
2. Click the event once to open the event form window
3. attempt to drag the event

### Context

Seems to be introduced in https://github.com/SwitchbackTech/compass/commit/c18b08cc4540e18313d3f0353671ee99c2c33310

that-one-arab:
@tyler-dane I discovered that the bug generally affects any event that has its form window open, so I refactored the issue title and description.

PR #302, opened 2025-03-13
🐛 Fix: dragging an event with its form window open does not work

that-one-arab:
## Description

Closes https://github.com/SwitchbackTech/compass/issues/298

I'm not sure why we have the below removed piece of code, but it was the culprit behind the issue.

I read the code comments about allowing drag only if form is open, but I couldn't comprehend it.

If it was part of an important UX logic we can revisit this in the future

#### Before
[Peek.webm](https://github.com/user-attachments/assets/6c5bd296-a940-42a8-8206-391058a5bbb7)

#### After
[Peek.webm](https://github.com/user-attachments/assets/65424633-4353-45a4-b7ad-cee010c762ed)

### DIFF
--- a/packages/web/src/views/Calendar/hooks/grid/useGridEventMouseDown.ts
+++ b/packages/web/src/views/Calendar/hooks/grid/useGridEventMouseDown.ts
@@ -95,10 +95,6 @@ export const useGridEventMouseDown = (
         if (timeoutId.current) {
           clearTimeout(timeoutId.current);
         }
-        if (isEventFormOpen()) {
-          cleanup(element, onMouseMove, onMouseUp);
-          return;
-        }
         onDrag(event);
         cleanup(element, onMouseMove, onMouseUp);
       }


tyler-dane (approved):

tyler-dane (approved):

tyler-dane:
This change broke this test:

Screenshot 2025-03-17 at 5 52 10 AM


The test validates a useful use-case; we don't want onDrag rendering superfluously, which will cause performance issues that are difficult to debug down the road. 

Can you investigate how to preserve the functionality while also not breaking the test? @that-one-arab

tyler-dane:
Also please look into why CI isn't running for your PRs. CI should run automatically on every push as part of the PR process, so I'm confused why this wasn't caught before merge 🤔

Screenshot 2025-03-17 at 5 56 09 AM

tyler-dane:
> > Also please look into why CI isn't running for your PRs. CI should run automatically on every push as part of the PR process, so I'm confused why this wasn't caught before merge 🤔
> > Look at your previous PR - they don't have the green checks, which means CI isn't running. @that-one-arab
> > Screenshot 2025-03-17 at 5 56 09 AM
> 
> My bad I wasn't aware of the CI checks, I'll make sure it runs in future PRs.

It should run automatically on every push based on `.github/workflows/test.yml`. If you ever find out why it hasn't been running for your PRs, LMK. Perhaps some setting/config is off

Issue #266, opened 2025-02-20
Grid continues scrolling after user finishes dragging


tyler-dane:
### Where did this happen?

Local code with an up-to-date main branch, Production app

### Expected Behavior

Given that a user has finished dragging an event by mousing up
And the grid needed to scroll during the process of the drag (in order to keep showing the event rectangle)
Then the scrolling should stop
And the user should be able to see their updated event
And the event should be in the correct location (based on where they moused up)

### Current Behavior

The scrolling continues until reaching the top/bottom

If the user stopped dragging during a scroll, the event is placed at the very top / bottom of the grid

https://github.com/user-attachments/assets/5399eefb-1d8a-4e41-bf75-5497ae998ae1

### Steps to Reproduce

1. Drag event up/down until the grid starts scrolling
2. Release the mouse during the scroll
3. Notice the event is not in the correct location
4. Notice that the scroll does not stop

### Context

This was introduced in #262

that-one-arab:
I linked the wrong issue to the PR, reopening this...

that-one-arab:
Part of the issue (indefinite scrolling) was fixed in https://github.com/SwitchbackTech/compass/pull/304, I will focus on fixing the second part of this issue (correctly setting the event to the right time duration when its scrolled up)

PR #307, opened 2025-03-17
Bug set right duration when dragging up outside maingrid view

that-one-arab:
## Description

Closes https://github.com/SwitchbackTech/compass/issues/266

The culprit was the deleted lines. I tested the deleted changes as thoroughly as possible and detected no noticable UX change. I am confused on why we needed these deleted lines, but as of now I don't see any need for them.

#### Before
[Peek 2025-03-17 21-16.webm](https://github.com/user-attachments/assets/383d1209-094d-42d9-b5a8-092aa6b1fdca)

#### After
[Peek.webm](https://github.com/user-attachments/assets/8451ec52-effc-493b-a918-b01253611361)

### DIFF
--- a/packages/web/src/views/Calendar/hooks/grid/useDateCalcs.ts
+++ b/packages/web/src/views/Calendar/hooks/grid/useDateCalcs.ts
@@ -27,12 +27,7 @@ export const useDateCalcs = (
   };
 
   const getDateByXY = (x: number, y: number, firstDayInView: Dayjs) => {
-    const isOverAllDayRow =
-      y > measurements.allDayRow.top && y < measurements.mainGrid.top;
-
     let date = getDateByX(x, firstDayInView);
-
-    if (isOverAllDayRow) return date;
 
     const minutes = getMinuteByY(y);
     date = date.add(minutes, "minutes");
--- a/packages/core/src/util/app.util.ts
+++ b/packages/core/src/util/app.util.ts
@@ -1,0 +1,10 @@
+import { NodeEnv } from "@core/constants/core.constants";
+import { isDev } from "@core/util/env.util";
+
+export const devAlert = (message: string) => {
+  if (isDev(process.env["NODE_ENV"] as NodeEnv)) {
+    alert(message);
+  } else {
+    console.warn(message);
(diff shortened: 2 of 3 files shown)


that-one-arab:
@tyler-dane I couldn't figure out why we needed the deleted lines. I hope I didn't miss anything. Can you take a look and let me know your thoughts?

packages/web/src/views/Calendar/hooks/grid/useDateCalcs.ts
```diff
@@ -27,13 +27,8 @@ export const useDateCalcs = (
...
   const getDateByXY = (x: number, y: number, firstDayInView: Dayjs) => {
-    const isOverAllDayRow =
-      y > measurements.allDayRow.top && y < measurements.mainGrid.top;
-
     let date = getDateByX(x, firstDayInView);
 
-    if (isOverAllDayRow) return date;
-
     const minutes = getMinuteByY(y);
     date = date.add(minutes, "minutes");
```
tyler-dane, line 34: Took a look at this and am also not sure why this would be necessary.

My best guess is that it was there because we needed a way to differentiate between getting the precise datetime for timed events and getting the date for all-day events. But now that we have `getDateByX` and `getDateByXY`, I don't think we'd need this extra check.

tyler-dane:
> @tyler-dane Also CI isn't running for this PR. Please let me know if you would like me to take a look or if you want to examine it yourself.

I think it's because you're pushing to a fork instead of this repo. Try using this repo, creating a branch, and pushing to that. This way, when you create a PR, the target will be `main` instead of your `[SwitchbackTech:main](https://github.com/SwitchbackTech/compass/tree/main)` (your fork)

Something like this

```
git remote add upstream git@github.com:SwitchbackTech/compass.git
git checkout -b test-branch
git push upstream test-branch
```

Then open a PR. You should see the workflow run on every push:

https://github.com/SwitchbackTech/compass/actions?query=event%3Apush

that-one-arab:
@tyler-dane Got it thank you for pointing this out. I will close this PR and re-create it

PR #312, opened 2025-03-18
🐛 Fix: set right duration when dragging up outside maingrid view

that-one-arab:
## Description

PR originally created in https://github.com/SwitchbackTech/compass/pull/307

Closes https://github.com/SwitchbackTech/compass/issues/266

The culprit was the deleted lines. I tested the deleted changes as thoroughly as possible and detected no noticable UX change. I am confused on why we needed these deleted lines, but as of now I don't see any need for them.

#### Before
[Peek 2025-03-17 21-16.webm](https://github.com/user-attachments/assets/383d1209-094d-42d9-b5a8-092aa6b1fdca)

#### After
[Peek.webm](https://github.com/user-attachments/assets/8451ec52-effc-493b-a918-b01253611361)

### DIFF
--- a/packages/core/src/util/app.util.ts
+++ b/packages/core/src/util/app.util.ts
@@ -1,0 +1,10 @@
+import { NodeEnv } from "@core/constants/core.constants";
+import { isDev } from "@core/util/env.util";
+
+export const devAlert = (message: string) => {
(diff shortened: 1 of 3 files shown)


that-one-arab:
Checks didn't pass to due changes introduced in https://github.com/SwitchbackTech/compass/pull/302#issuecomment-2729070008

Issue #186, opened 2024-12-13
Events lose position during week change


tyler-dane:
## Prerequisites

- [x] Using an up-to-date main branch

## Current Behavior

Look at the all-day row during week changes:

https://github.com/user-attachments/assets/bfe66015-3211-47a2-b261-2b9274c05093

## Steps to Reproduce

1. Add an all-day and timed event in a different week.
1. Move the view to that week, so you can see the events. They should render properly.
1. Move back to the previous week.
1. Notice: that an all-day event with the duration of the whole view quickly shows (for less than a second) then disappears.
1. Notice: timed events shift to the left, losing their original position 

## Possible Solution (Not obligatory)

**Non-ideal Fix**
One approach would be to hide the issue from the user. This would involve adding an overlay over the calendar so the user doesn't see the events changing position. This overlay could be a loading spinner, or an animated skeleton. 

This approach is non-ideal because it doesn't fix the root cause and decreases performance by requiring the UI to render an overlay and the grid 

**Ideal Fix**
Another approach is to find a way to prevent this from happening in the first place. This would require us to skip the computation of an event's width/height/grid coordinates during a pending navigation.

### Root Cause
When we change the week, `CalendarView` re-renders, triggering re-renders in all its component tree until reaching event components, which re-calculates their position and size. 

## Context
This makes the app feel less smooth and lowers the user's confidence in its performance and reliability.

PR #337, opened 2025-04-03
🐛 Fix: grid events losing position when changing to next/prev week

that-one-arab:
## Description
Closes https://github.com/SwitchbackTech/compass/issues/186

disable rendering events when a backend request is occurring, that way we address the issue where events lose position whenever we switch current week.

#### Before
https://github.com/user-attachments/assets/7867dad7-aed1-4ce1-a6c3-30c0881056b5

#### After
https://github.com/user-attachments/assets/b10253be-0772-45ec-9d04-234290431409


that-one-arab:
Reopened here https://github.com/SwitchbackTech/compass/pull/338

PR #338, opened 2025-04-03
🐛 Fix: grid events losing position when changing to next/prev week

that-one-arab:
## Description
Closes https://github.com/SwitchbackTech/compass/issues/186

disable rendering events when a backend request is occurring, that way we address the issue where events lose position whenever we switch current week.

#### Before
https://github.com/user-attachments/assets/7867dad7-aed1-4ce1-a6c3-30c0881056b5

#### After
https://github.com/user-attachments/assets/b10253be-0772-45ec-9d04-234290431409

### DIFF
--- a/packages/web/src/ducks/events/selectors/util.selectors.ts
+++ b/packages/web/src/ducks/events/selectors/util.selectors.ts
@@ -10,6 +10,9 @@ export const selectIsProcessing = (state: RootState) =>
   isProcessing(state.events.createEvent) ||
   isProcessing(state.events.getWeekEvents);
 
+export const selectIsGetWeekEventsProcessing = (state: RootState) =>
+  isProcessing(state.events.getWeekEvents);
+
 export const selectEventIdsBySectionType = (
   state: RootState,
   type: SectionType,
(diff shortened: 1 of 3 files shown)


packages/web/src/common/store/helpers/index.ts
```diff
@@ -44,6 +44,12 @@ export const createAsyncSlice = <
 
   const reducers = {
     request: (state, _action: PayloadAction<Draft<RequestPayload>>) => {
+      // When running tests, we use mock data, so we don't need to fetch the week events from the API
+      // See comments in https://github.com/SwitchbackTech/compass/pull/338 for dev notes
+      if (process.env.NODE_ENV === "test") {
+        return;
+      }
```
that-one-arab, line 51: placing the condition check here (reducers.request()) versus placing the condition where the method is being called has its pros and cons, I opted to place it here because:
- It is a quicker solution
- It hides this piece of logic inside the function, abstracting the method when viewed from outside

Some cons would be:
- May have cases where this logic is harder to debug because it is abstracted/hidden

Issue #420, opened 2025-05-16
Add `f` shortcut to edit header note


tyler-dane:
### Feature Description
As a user, when I press the `f` key, I want to start editing the header note, so that I can change it easily.

I also want to edit it from the `META` + `k` shortcut, so I can edit the header note when I forget what the shortcut is

Make it easier to edit the header note with a shortcut or command palette action.

`f` key should be used, as it's related to the work 'focus', which is used as in placeholder text for the header note. The `f` is also where the left index finger is located on US keyboards, which makes it very easy to press.

<img width="1915" alt="Image" src="https://github.com/user-attachments/assets/7b9e0d52-7dca-47e2-8964-f550164104d6" />

<img width="1706" alt="Image" src="https://github.com/user-attachments/assets/668a80b5-badc-4522-ad9c-a83dfd01aabb" />

### Use Case

This'll make it easier to edit the note frequently 

### Additional Context

Acceptance Criteria
- [ ] Tooltip renders when hovering over header note
- [ ] Shortcut works on all major browsers and operating systems
- [ ] Shortcut is added as an action to command palette.

sarans-h:
@tyler-dane pls share the screenshot and assign the issue to me please

tyler-dane:
Hey @sarans-h , added screenshots and assigned it to you

tyler-dane:
@sarans-h Based on you closing the other PR and inactivity, I unassigned you and moved this back to Ready. Happy to give it back to you if you want to give it another shot.

pheobeayo:
I am a Web3 front-end developer skilled in technical documentation and software development. I greatly understand technologies like JavaScript, Typescript, Cairo, Rust and Solidity. I have contributed successfully to open-source projects like Stark Quest among others. 
My Approach includes:
- Study the code styles implemented.
- Implement the "f" shortcut to edit header note.
- Test the code
- Create a Pull Request (PR)
- Request for a Review
- Make changes based on the Review of the PR

tyler-dane:
@pheobeayo I assigned this to you and added an estimated end date. Please LMK if you'll need more time

pheobeayo:
@tyler-dane Okay, thank you. I will get right on it.

tyler-dane:
Unassigning and moving back to Ready due to inactivity

pheobeayo:
> Unassigning and moving back to Ready due to inactivity

[...]

PR #430, opened 2025-05-19
Feat/issue#420 added f shortcut and tooltip

sarans-h:
fixing issue #420 
1. Added f shortcut
2. Added in cmd pallet
3. added tooltip 
image
image

### DIFF
--- a/packages/web/src/views/Calendar/Calendar.tsx
+++ b/packages/web/src/views/Calendar/Calendar.tsx
@@ -69,15 +69,17 @@ export const CalendarView = () => {
         isSidebarOpen={isSidebarOpen}
       >
         <SidebarDraftProvider dateCalcs={dateCalcs} measurements={measurements}>
-          <Draft measurements={measurements} weekProps={weekProps} />
-          {isSidebarOpen && (
-            <Sidebar
-              dateCalcs={dateCalcs}
-              measurements={measurements}
-              weekProps={weekProps}
-              gridRefs={gridRefs}
-            />
-          )}
+          <ContextMenuWrapper>
+            <Draft measurements={measurements} weekProps={weekProps} />
+            {isSidebarOpen && (
+              <Sidebar
+                dateCalcs={dateCalcs}
+                measurements={measurements}
+                weekProps={weekProps}
+                gridRefs={gridRefs}
+              />
+            )}
+          </ContextMenuWrapper>
         </SidebarDraftProvider>
         <StyledCalendar direction={FlexDirections.COLUMN} id={ID_MAIN}>
           <Header
--- a/packages/web/src/common/utils/event.util.ts
+++ b/packages/web/src/common/utils/event.util.ts
@@ -237,9 +237,29 @@ export const isEventInRange = (
   return isStartDateInRange || isEndDateInRange;
 };
 
-export const isOptimisticEvent = (event: Schema_GridEvent) => {
+export const isOptimisticEvent = (event: Schema_Event) => {
   const isOptimistic = event._id?.startsWith(ID_OPTIMISTIC_PREFIX) || false;
   return isOptimistic;
+};
+
+export const getSomedayEventCategory = (
+  event: Schema_Event,
+): Categories_Event.SOMEDAY_MONTH | Categories_Event.SOMEDAY_WEEK => {
+  if (!event.isSomeday) {
+    throw new Error(
+      `Event is not a someday event. Event: ${JSON.stringify(event)}`,
+    );
+  }
+
+  const startDate = dayjs(event.startDate);
+  const endDate = dayjs(event.endDate);
+
+  const diffInDays = endDate.diff(startDate, "day");
+
+  if (diffInDays > 7) {
+    return Categories_Event.SOMEDAY_MONTH;
+  }
+  return Categories_Event.SOMEDAY_WEEK;
 };
 
 export const prepEvtAfterDraftDrop = (
--- a/packages/web/src/components/ContextMenu/ContextMenuItems.tsx
+++ b/packages/web/src/components/ContextMenu/ContextMenuItems.tsx
@@ -3,9 +3,13 @@ import { Copy, PenNib, Trash } from "@phosphor-icons/react";
 import { Priorities } from "@core/constants/core.constants";
 import { colorByPriority } from "@web/common/styles/theme.util";
 import { Schema_GridEvent } from "@web/common/types/web.event.types";
-import { assembleGridEvent } from "@web/common/utils/event.util";
+import {
+  assembleGridEvent,
+  getSomedayEventCategory,
+} from "@web/common/utils/event.util";
 import IconButton from "@web/components/IconButton/IconButton";
 import { useDraftContext } from "@web/views/Calendar/components/Draft/context/useDraftContext";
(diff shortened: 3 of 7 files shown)


packages/web/src/views/Calendar/components/Header/HeaderNote/HeaderNote.tsx
```diff
@@ -134,34 +146,37 @@ export const HeaderNote = () => {
...
+          range.selectNodeContents(focusRef.current);
+          range.collapse(false);
+          sel?.removeAllRanges();
+          sel?.addRange(range);
+        }
+      }, 0);
     }
-  }, [isEditing]);
+    prevEditingRef.current = editing;
+  }, [editing, note]);
```
tyler-dane: This is a creative solution...but it's also complex and difficult to understand for anyone else who tries to read this code. I'm also concerned about overrelying on range, which could easily break as the UI changes.

tyler-dane:
Basic functionality is looking good, but I have a few QA concerns.

Whenever trigger the shortcut or clicking the header note intput, the someday events are refetched. This tells me there is some dependency between the two being introduced. I'm guessing it has to do with the shortcut handler, the `useEffect`, or the new props.

https://github.com/user-attachments/assets/0a54df78-c816-4536-ab05-cb8ca95588d0

packages/web/src/views/Calendar/components/Header/Header.tsx
```diff
@@ -67,7 +74,10 @@ export const Header: FC<Props> = ({ scrollUtil, today, weekProps }) => {
             />
           </TooltipWrapper>
         </StyledLeftGroup>
-        <HeaderNote />
+        <HeaderNote
+          isEditing={isEditingHeaderNote}
+          setIsEditing={setIsEditingHeaderNote}
+        />
```
tyler-dane: passing around a setter like this isn't ideal, as it's going to open us up to bugs and unnecessary dependencies.

tyler-dane (changes requested):
Requesting changes based on other comments.

packages/web/src/views/Calendar/Calendar.tsx
```diff
@@ -24,24 +24,22 @@ import { Styled, StyledCalendar } from "./styled";
...
   const scrollUtil = useScroll(gridRefs.mainGridRef);
-
   const dateCalcs = useDateCalcs(measurements, gridRefs.mainGridRef);
-
   const isCurrentWeek = weekProps.component.isCurrentWeek;
   const util = weekProps.util;
 
+  // Add editing state for HeaderNote
+  const [isEditingHeaderNote, setIsEditingHeaderNote] = useState(false);
+  const onEditHeaderNote = () => setIsEditingHeaderNote(true);
```
that-one-arab: The `Calendar` component is the entry point to the app. This component should be abstracted as much possible and its implementation details should be hidden as much as possible, so exposing this implementation logic here is not ideal and would introduce tech debt.

packages/web/src/views/Calendar/Calendar.tsx
```diff
@@ -85,6 +83,8 @@ export const CalendarView = () => {
             scrollUtil={scrollUtil}
             today={today}
             weekProps={weekProps}
+            isEditingHeaderNote={isEditingHeaderNote}
+            setIsEditingHeaderNote={setIsEditingHeaderNote}
```
that-one-arab: Prefer not to pass around props like this as it couples the component to these props, prefer to implement separation of concerns.

packages/web/src/views/Calendar/components/Header/HeaderNote/HeaderNote.tsx
```diff
@@ -21,9 +23,19 @@ type CursorPosition = {
...
+  isEditing,
+  setIsEditing,
+}: {
+  isEditing?: boolean;
+  setIsEditing?: (v: boolean) => void;
+}) => {
   const [note, setNote] = useState("");
-  const [isEditing, setIsEditing] = useState(false);
+  const [internalEditing, setInternalEditing] = useState(false);
+  const editing = typeof isEditing === "boolean" ? isEditing : internalEditing;
```
that-one-arab: I am seeing this `typeof variable === "type"` pattern in this PR in a few places and I believe it is unnecessary as developers would generally not make a mistake of passing incorrect type to the variable so no need to explicitely check the type. You can just shorten it to `if (!!variable)` or `if (variable)`

packages/web/src/views/Calendar/components/Header/HeaderNote/HeaderNote.tsx
```diff
@@ -263,7 +278,14 @@ export const HeaderNote = () => {
 
   return (
     <StyledNoteContainer>
-      {isEditing ? (
+     <TooltipWrapper
+        description={
+        editing ? "Press Enter to save" : "Click to edit header note"
+      }           
+      shortcut={editing ? "Enter" : "F"}
+      onClick={handleNoteClick}
```
that-one-arab: I believe this would trigger 2 `handleNoteClick` calls instead of 1, because this callback is also called below. Please check that this is not the case.

packages/web/src/views/Calendar/hooks/shortcuts/useShortcuts.ts
```diff
@@ -141,11 +143,23 @@ export const useShortcuts = ({
         [Key.Z]: () => {
           navigate(ROOT_ROUTES.LOGOUT);
         },
+        [Key.F]: () => {
+          if (
+            typeof onEditHeaderNote === "function" &&
+            document.activeElement &&
+            !(document.activeElement instanceof HTMLInputElement) &&
+            !(document.activeElement instanceof HTMLTextAreaElement) &&
+            document.activeElement.getAttribute("contenteditable") !== "true"
```
that-one-arab: I'm assuming you are trying to retrieve the active element and checking if the active element is the header note text, if yes, then this is an overly complicated way to do it and is open to bugs. Why not just check the `id` of the field?  (EG: `document.activeElement?.id === 'headerNoteInput'`)

packages/web/src/views/Calendar/hooks/shortcuts/useShortcuts.ts
```diff
@@ -167,5 +181,6 @@ export const useShortcuts = ({
     scrollUtil,
     util,
     tab,
+    onEditHeaderNote, // Add to deps
```
that-one-arab: There are lots of these type of comments, I'm assuming they are leftover by AI. It is fine to use AI generated code but please double check the code and groom it to ensure no such comments are left in the commits.

that-one-arab (changes requested):
I agree to everything @tyler-dane mentioned, and I left additional feedback that may prove helpful.

Sadly this PR introduces too many tech debt than its worth.

If you feel like cleaning up this PR would be too difficult, let me know! I will hook you up with an easier issue :)

packages/web/src/views/Calendar/hooks/shortcuts/useShortcuts.ts
sarans-h: yes i used ai for adding comments so that the reviewer must know what i did 
will update the pr as discussed above

that-one-arab: @sarans-h You don't need to set the comments inside the code itself. You can add your comments here inside the github PR, same way we are communicating now.

Issue #428, opened 2025-05-18
Creating an all day event at the start of the week does not display it after refreshing the page


that-one-arab:
### Expected Behavior

When creating an all day event for date at the start of the week  and then refreshing the page, the event should be visible

### Current Behavior

The event disapears as if I never created it in the first place

### Steps to Reproduce

1. Click on all day area for start of the week to create an all day event
2. Create the all day event
3. Refresh the page
4. Observe that the event is not loaded again

https://github.com/user-attachments/assets/ab34da5b-6591-440c-bc07-e2fe62ecd8b5

tyler-dane:
Since this is happening on the first day of the week and the event is saved to GCal, I'm thinking this is due to either the query on the backend or how the frontend is filtering the query.

that-one-arab:
@tyler-dane So I'm investigating and resolving this in https://github.com/SwitchbackTech/compass/pull/440

The issue is mainly a backend one, and has to do with all day events being stored in raw date string format while timed events are stored in ISO 8601 format.

I am prototyping a solution that I trust will resolve this problem in a smart and an efficient way, but I will need an additional day to do this, so I'll update the end date of the issue to the 23rd if that's okay with you.

tyler-dane:
Sounds good. Please just make sure to thoroughly unit test when making
backend changes.

tyler-dane:
Thinking through the root cause a little more, I agree with what you said in #440  about this being related to timezones. Google Calendar API includes a timezone property for every event. The best path forward is probably not to change the date format for all-day events, but to also add a timezone to our event model and use it in our query to catch events that are at the week's edge.

Un-assigning and putting this back to Ready. Me or the backend engineer can pick it up again once it's a higher priority.

PR #440, opened 2025-05-21
🐛 Fix: allday events not displaying if start-end date is start of week

that-one-arab:
## Description

Any all day event that is created at the start of the week (regardless of whether it was created on 2025-05-18 or not) is not displayed upon refreshing page.

## Context
We fire a single API request that returns both grid events and allday events in one go.

The API request that is fired sends a `start` and `end` parameters to specify the date range for the returned events. These parameters are an ISO 8601 date strings.

Timed events are stored as ISO 8601 date strings.

While all-day events are stored as raw date strings (`YYYY-MM-DD`)

## Problem
If we have the following scenario:
- Start of the week is **2025-02-02**
    - This means that the `start` and `end` parameters fired would be `2025-02-02T00:00:00+03:00` and `2025-02-08T23:59:59+03:00` (remember, we send them as ISO 8601 strings)
- We have a timed event who's start date is `2025-02-02T00:00:00+03:00`
- We have an all-day event who's start date is `2025-02-02`

The result would be:
- The timed event is returned, because timed event's start date `2025-02-02T00:00:00+03:00` can be lexicographically compared to the provided ISO 8601 date string
- The all-day event is **not returned**, because all-day event's start date `2025-02-02` cannot lexicographically compared to the provided ISO 8601 date string.

## Solution

- On the frontend side: ensure submitted dates conform to ISO 8601 standards. This will standardize saved event dates in the DB
- On the backend side: migrate event dates that do not conform to ISO 8601 standards to dates that do conform (we achieve that by inferring the timezone from previously saved timed events, if no timed events exist, we assign a default timezone (this is the only downside of this approach, but I believe it is necessary and should cause only minimal harm)), fix few tests.

We need to run `migrate-allday-events`  on staging and prod once for each env after merging this PR.


tyler-dane:
Closing after deciding to tackle problem in a different way (see #428 )

PR #453, opened 2025-05-25
🐛 Fix: allday events not displaying if start-end date is start of week (Attempt 2)

that-one-arab:
## Description
Follow up to https://github.com/SwitchbackTech/compass/pull/440

This PR solves the issue by aiming at solving it from the frontend side.

It resolves it by sending an additional HTTP request specifically for all day events. This works because the additional HTTP request is sent under date filters that are different than the get events HTTP request (uses YYYY-MM-DD format rather than ISO 8601)

Sadly, this introduces an additional HTTP request to be sent, which would degrade performance. This is the only way to resolve this issue solely from the frontend side that I know of.

This could be further optimized by implementing a backend side solution to send 2 database requests in 1 HTTP request, rather than sending 2 database requests in 2 HTTP requests, but since the goal is to implement a frontend solution only, I omitted doing that.

### DIFF
--- a/packages/backend/src/event/services/event.find.test.ts
+++ b/packages/backend/src/event/services/event.find.test.ts
@@ -90,7 +90,7 @@ describe("Jan 2022: Many Formats", () => {
       const filter = getReadAllFilter("user1", {
         // make sure these match text data exactly
         start: "2022-01-01T00:00:00+03:00",
-        end: "2022-01-020T11:11:11+03:00",
+        end: "2022-01-20T11:11:11+03:00",
       });
 
       const result = await eventCollection.find(filter).toArray();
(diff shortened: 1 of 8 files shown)


tyler-dane:
I appreciate the persistent, but this unfortunately also isn't going to work. The problem is that we're already taking a big performance hit due to #456. After testing this branch, I found the UX to be too slow, and I'm afraid the users will think the same and drop.

The bug this attempts to fix is not ideal, but it does save the event to the user's google calendar and our Compass DB. So, it results in some annoyance if a user makes an all day event on Sunday. But introducing this PR without fixing #456 would guarantee a worse UX for all users whenever they change weeks.

Thinking through the root cause a little more, I agree with what you said in #440 about this being related to timezones. Google Calendar API includes a timezone property for every event. The best path forward is probably not to change the date format for all-day events, but to also add a timezone to our `event` model and use it in our query to catch events that are at the week's edge.

We've spent enough time on this issue for now, so let's revisit after finishing #456 and figuring out how to change the event model on the backend.

PR #465, opened 2025-05-28
fix: change duplicate event shortcut from CTRL+d to META+d (#446)

### LINKED ISSUE #446: `META + d` does not duplicate event
### Where did this happen?

Production app, Local code with an up-to-date main branch

### Expected Behavior

Duplicated when focused on an event and typing `META + d`

This is the default "create bookmark" shortcut in Chrome. However, we can override it using event listeners and event functions so that it doesn't trigger.

```ts
window.addEventListener("keydown", function (e) {
  if (e.metaKey && e.key.toLowerCase() === "d") {
    e.preventDefault(); // prevent browser bookmark
    e.stopPropagation(); // stop it from bubbling up
    // trigger your duplication logic here
    console.log("Duplicate event shortcut triggered");
  }
});
```

We should stick to `META` for consistency. This'll require us to prevent the default bookmark shortcut from trigger, like I shared in my example code

https://github.com/user-attachments/assets/3b2bb063-4405-4018-a2a4-5e97f91cc92b

### Current Behavior

`CTRL + d` duplicates an event

### Steps to Reproduce

1. Create an event (all-day or timed)
2. Click on it
3. Press `META + d`
4. Event should be duplicated, and the create bookmark event should not trigger
5. Repeat
6. Press `CTRL + d`
7. Nothing should happen

### Possible Solution

Change the handler and the conditional that checks for the key.

### Context

Bug was introduced in #418

akintewe:
Problem
- META + d was creating browser bookmarks instead of duplicating events
- CTRL + d was duplicating events (inconsistent with other META shortcuts)
- Browser's default bookmark shortcut (META + d) was interfering with app functionality

Solution
- ✅ Changed duplicate event hotkey from `"ctrl+d"` to `"meta+d"`
- ✅ Existing `preventDefault()` logic stops browser bookmark creation
- ✅ Added comprehensive test coverage for META + d keyboard shortcut
- ✅ Maintains consistency with other META key shortcuts in the app

Testing
- ✅ Unit test verifies META + d calls `onDuplicate` correctly
- ✅ Test confirms no unintended side effects (other functions not called)
- ✅ Existing `handleIgnoredKeys` prevents browser default behavior

Changes
- `packages/web/src/views/Forms/EventForm/EventForm.tsx`: Updated hotkey handler
- `packages/web/src/views/Forms/EventForm/EventForm.test.tsx`: Added test coverage

Closes #446

### DIFF
--- a/packages/web/src/views/Forms/EventForm/EventForm.test.tsx
+++ b/packages/web/src/views/Forms/EventForm/EventForm.test.tsx
@@ -85,6 +85,52 @@ test("should call onConvert when meta+< (meta + shift + comma) keyboard shortcut
   expect(mockOnDelete).not.toHaveBeenCalled();
 });
 
+test("should call onDuplicate when meta+d keyboard shortcut is used", async () => {
+  const sampleEvent: Schema_Event = {
+    _id: "event123",
+    title: "Test Event for Duplication",
+    startDate: "2025-04-10",
+    endDate: "2025-04-10",
+    isAllDay: false,
+  };
+  const mockOnClose = jest.fn();
+  const mockOnConvert = jest.fn();
+  const mockOnSubmit = jest.fn();
+  const mockSetEvent = jest.fn();
+  const mockOnDelete = jest.fn();
+  const mockOnDuplicate = jest.fn();
+
+  render(
+    <div>
+      <EventForm
+        event={sampleEvent}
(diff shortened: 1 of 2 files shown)


that-one-arab (approved):
Looks good! Thanks!

PR #471, opened 2025-05-30
🐛 Fix: Implement hotfix to address events start of week date filter not being returned consistently

that-one-arab:

This hotfix resolves the issue by adjusting start date to be 1 day before the actual start date, this allows us to extend our date range to cover additional events that may not be returned. We ensure that the returned events match the initial date range by manually filtering them.


tyler-dane:
I'm good with this approach.

Wondering how it'll work if there are events on Saturday, though. Would those show up on Sunday now?

As long as weird bugs like that aren't introduced and the tests pass, I think we can merge this one

that-one-arab:
@tyler-dane Fixed failing tests and added videos to better show results. LMK your thoughts.

that-one-arab:
@tyler-dane I guess you missed my notification. i'm going ahead with the merge for now.

PR #472, opened 2025-05-30
Refactor keyboard event handling

oludaara:
use single event listener for EventForm and SomedayEventForm
This PR refactors the EventForm and SomedayEventForm components to optimize keyboard shortcut handling by:

Removing multiple useHotkeys and individual input onKeyDown listeners.
Centralizing all keyboard event handling in a single form-level onKeyDown event listener.
Maintaining all existing keyboard shortcut functionality (e.g., CMD/CTRL+Enter to submit, CMD+D to duplicate, CMD+Shift+Comma to convert, Delete key to delete, etc.).
Improving performance and maintainability by reducing the number of event listeners in the DOM.
Keeping all logic type-safe and clearly organized.

### DIFF
--- a/packages/web/src/views/Forms/EventForm/EventForm.tsx
+++ b/packages/web/src/views/Forms/EventForm/EventForm.tsx
@@ -1,6 +1,5 @@
 import dayjs from "dayjs";
-import React, { useCallback, useEffect, useRef, useState } from "react";
-import { Key } from "ts-key-enum";
+import React, { useEffect, useRef, useState } from "react";
 import { Priorities } from "@core/constants/core.constants";
 import { ID_EVENT_FORM } from "@web/common/constants/web.constants";
 import {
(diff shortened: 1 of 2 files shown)


oludaara:
AM done with it can you check and notify if any issue

that-one-arab (changes requested):
Thanks for the PR, but sadly it has some issues:
- `META + Enter` does not work on Mac, atleast
- Some tests are failing
- Linting issues

Also, I don't see any good reason for removing `useHotKeys` functionality, this library has always been dependable. Could you emphasize on why we needed to remove it?

Good luck and let me know if you need anything!

that-one-arab:
@oludaara Yea that approach sounds good!
But tests are still failing unfortunately, I think it has something to do with the custom hook you built not being inside the branch when you pushed it. 
Can you please fix that then ping me again?

that-one-arab:
@oludaara You can find it here
https://github.com/SwitchbackTech/compass/blob/main/packages/backend/.env.example

oludaara:
i just have to push these code now i have really tried my best but i dont know why i can run the code locally tried all fix but still the same seems like a backend issue or something checked the .env all is still using placeholder values that need to be replaced with actual configuration. Am tired ill push the code for the meta + enter if it works fine and if it doesnt shame on me.......

Thanks for you time now i know my limitations

oludaara:
My laptop died I'll push it when the electricity is back

PR #475, opened 2025-06-01
✨ feat(toast): prevent duplicate toasts

### LINKED ISSUE #445: Replace alert with better UX when attempting to edit a recurring event
### Feature Description

As a user, I want it to be clear that I can't edit a recurring event when I try to do so. But I also don't want the alert notification to be obnoxious, so that I can retain my focus.

Some ideas for a better UX:
- Make the recurring events darker (using the same colors as the events in the past). This will help prevent the problem of the user trying to make an edit in the first case.
- Replace the regular `alert()` with a toast (use `react-toastify`, which we already use in the `SomedayEventForm`)

### Use Case

This'll help users still see all their events while also making it less punitive when they naturally attempt to edit a recurring event.

### Additional Context

We currently only support viewing recurring events in Compass; we do not support making a change to a recurring event (yet). 

The current UX:

Image

7-falseparfait:
Replaces the intrusive browser alert() with a non-blocking toast notification when a user attempts to edit a recurring event. This enhances UX and keeps interactions smoother.

Closes #445 .

### DIFF
--- a/package.json
+++ b/package.json
@@ -29,7 +29,8 @@
     "test": "yarn jest",
     "test:backend": "yarn jest backend",
     "test:core": "yarn test core",
-    "test:web": "yarn test web"
+    "test:web": "yarn test web",
+    "lint": "eslint . --ext .js,.jsx,.ts,.tsx"
   },
   "dependencies": {
     "@compass/backend": "*",
--- a/package.json
+++ b/package.json
@@ -29,8 +29,7 @@
     "test": "yarn jest",
     "test:backend": "yarn jest backend",
     "test:core": "yarn test core",
-    "test:web": "yarn test web",
-    "lint": "eslint . --ext .js,.jsx,.ts,.tsx"
+    "test:web": "yarn test web"
   },
   "dependencies": {
     "@compass/backend": "*",
--- a/packages/web/src/common/utils/event.util.ts
+++ b/packages/web/src/common/utils/event.util.ts
@@ -22,6 +22,7 @@ import {
   Schema_OptimisticEvent,
   Schema_SomedayEventsColumn,
 } from "../types/web.event.types";
+import { toastWithoutDuplication } from "./toast";
 
 dayjs.extend(isSameOrAfter);
 dayjs.extend(isSameOrBefore);
@@ -210,7 +211,9 @@ export const handleError = (error: Error) => {
   }
 
   if (code === Status.INTERNAL_SERVER) {
-    alert("Something went wrong behind the scenes. Please try again later.");
+    toastWithoutDuplication(
+      "Something went wrong behind the scenes. Please try again later.",
+    );
     window.location.reload();
   }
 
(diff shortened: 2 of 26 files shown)


package.json
```diff
@@ -29,7 +29,8 @@
     "test": "yarn jest",
     "test:backend": "yarn jest backend",
     "test:core": "yarn test core",
-    "test:web": "yarn test web"
+    "test:web": "yarn test web",
+    "lint": "eslint . --ext .js,.jsx,.ts,.tsx"
```
that-one-arab: modifiying `package.json` is outside the scope of this PR

packages/web/src/views/Calendar/components/Event/Grid/GridEvent/GridEvent.tsx
```diff
@@ -6,6 +6,7 @@ import React, {
   memo,
   useMemo,
 } from "react";
+import { toast } from "react-toastify";
```
that-one-arab: Are these imports actually used?

packages/web/src/common/utils/event.util.ts
```diff
@@ -22,6 +22,7 @@ import {
   Schema_OptimisticEvent,
   Schema_SomedayEventsColumn,
 } from "../types/web.event.types";
+import { toastWithoutDuplication } from "./toast";
```
that-one-arab, line 25: Can you elaborate more on the issue regarding duplicate toasts? I read the linked issue and I didn't find anything related to it

7-falseparfait: Sure! I created a small wrapper around the default toast function because I noticed that when certain events are triggered multiple times quickly, it can flood the UI with duplicate toasts. This function reduces that by preventing repeated toasts from showing up too often. It takes the same arguments as the regular toast, but improves the user experience by avoiding clutter.

packages/web/src/views/Calendar/components/Event/Grid/GridEvent/GridEvent.tsx
7-falseparfait: You're right the toast import isn’t used directly anymore since we're using the toastWithoutDuplication abstraction instead. I’ve removed the unused import in the latest commit. Thanks for pointing it out!

packages/web/src/common/utils/event.util.ts
that-one-arab: @7-falseparfait Can you elaborate on which certain events trigger toast duplication?
In some cases we should be okay with duplicating toast notifications (EG: user saved multiple events rapidly)

7-falseparfait: @that-one-arab Sure! The main case  aimed to avoid was repeated identical toast messages for the "Can't edit recurring events" scenario. For example, if a user clicks on multiple recurring events quickly, toastWithoutDuplication helps prevent stacking the same error message multiple times.
Overall, the intention is to clearly inform the user that they can't edit a recurring event when they try to do so without the alert notification becoming too distracting or disruptive to their focus.

that-one-arab:
Closing since this is an outdated attempt and new PR appears to be here https://github.com/SwitchbackTech/compass/pull/491

PR #479, opened 2025-06-02
Fixed the META + ENTER  issue

oludaara:


that-one-arab (changes requested):
Thanks for the PR! I have the following concerns:
- Your implemented `useFormKeyboardSubmit` hook listens to the document object and requires a lot of boilerplate code to be properly setup (needs an external ref object, need to conditionally check if the ref is open, etc...). These changes make me question how performant this implementation actually is compared to `react-form-hook` (based on our previous conversation in https://github.com/SwitchbackTech/compass/pull/472
- We now have 2 separate implementations to handle keyboard shortcuts. This may create confusion and/or inconsistincies in the future for devs working on the project.
- You have changes in your package-lock, yarn.lock and tsconfig that should not be commited.
- You have a lot of changes related to TS interfaces. These changes make the PR significantly harder to review and test. Please keep them organized in separate commits or in a separate PR in the future

I advise you to stick to a working solution using `react-form-hook`

PR #490, opened 2025-06-07
🐛 Fix: Someday events categorization logic

that-one-arab:


packages/web/src/common/types/web.event.types.ts
```diff
@@ -28,6 +28,12 @@ export interface Schema_GridEvent extends Schema_Event {
   };
 }
```
tyler-dane, line 36: Can you use zod to create a schema and then infer the type? We do that on the backend, and I find that it makes it easier to validate data and keep types consistent.

that-one-arab: @tyler-dane great idea thanks for the headsup

that-one-arab: @tyler-dane Noticed we started having a few failing tests. Will take an additional day on this to investigate since I believe it is critical to have functional tests due to me focusing on someday events this month

packages/web/src/common/types/web.event.types.ts
```diff
@@ -28,6 +34,12 @@ export interface Schema_GridEvent extends Schema_Event {
...
 
+export interface Schema_SomedayGridEvent extends Schema_Event {
+  _id: string;
+  order: number;
+  isSomeday: true;
+}
+
```
tyler-dane, line 42: It looks like this is manually creating an interface that matches the `SomedayEventSchema`.  Instead of doing this manually and risking it drifting as the schema changes, you can use `z.infer` to create the type from the schema 

@that-one-arab

that-one-arab: @tyler-dane Ah thanks for the tip, will do that in the current PR I'm working on

PR #491, opened 2025-06-07
✨feat: add preventDuplicate flag for toast, scoped to recurring events

### LINKED ISSUE #445: Replace alert with better UX when attempting to edit a recurring event

7-falseparfait:
This supersedes previous branches  
closes #445

### DIFF
--- a/packages/web/src/views/Calendar/components/Event/styled.ts
+++ b/packages/web/src/views/Calendar/components/Event/styled.ts
@@ -24,10 +24,15 @@ interface StyledEventProps {
   priority: Priority;
   top: number;
   width: number;
+  isRecurring: boolean;
 }
 
 export const StyledEvent = styled.div.attrs<StyledEventProps>((props) => {
   const getBgColor = () => {
+    if (props.isRecurring || props.isInPast) {
+      return darken(colorByPriority[props.priority], 50);
+    }
+
     if (props.isResizing || props.isDragging) {
       return brighten(colorByPriority[props.priority]);
     }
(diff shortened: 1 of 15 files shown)


packages/web/src/views/Calendar/components/Event/styled.ts
```diff
@@ -24,10 +24,15 @@ interface StyledEventProps {
...
   top: number;
   width: number;
+  isRecurring: boolean;
 }
 
 export const StyledEvent = styled.div.attrs<StyledEventProps>((props) => {
   const getBgColor = () => {
+    if (props.isRecurring || props.isInPast) {
+      return darken(colorByPriority[props.priority], 50);
+    }
```
that-one-arab: This makes a past or recurring event **way too dark**

Here is an example of how to do it correctly 
image

that-one-arab (changes requested):
Looks good! Just one more small change and we're good to go.

that-one-arab:
@7-falseparfait looks good now but your changes are making a test fail I believe.

More specifically, this one: `SomedayEventForm Hotkeys › should trigger delete flow when delete keyboard shortcut is used and confirmed`

Can you please take a look?

that-one-arab:
@7-falseparfait Closing due to inactivity. Please feel free to re-open after resuming work.

PR #476, opened 2025-06-01
✨ Feat: Added shortcut for all day event and added it to the command …

chiefDevOvrlrd:
added a key as a  shortcut. added the shortcut to the command palette

PR #494, opened 2025-06-08
 ✨ Feat: Added shortcut for all day event and added it to the command...

chiefDevOvrlrd:
added a key as a shortcut. added the shortcut to the command palette
issue #476


chiefDevOvrlrd:
@that-one-arab

PR #499, opened 2025-06-09
Fix: Submitting the event form with Meta+Enter

### LINKED ISSUE #458: Event forms do not submit when typing `META`+ `ENTER`
### Where did this happen?

Local code with an up-to-date main branch, Production app

### Expected Behavior

When typing `META + ENTER` anytime the event form is open, the form should submit.

It should not matter if the focus is on the title, the priority buttons, the repeat button, or the description textfield. 

### Current Behavior

`META + ENTER` only works when the focus is on the title or description. It does not submit the form when the focus is on the priority buttons or repeat button.

https://github.com/user-attachments/assets/ae041894-6fc7-464b-99aa-cd085c1d4d23

**Acceptance Criteria**
- [ ] Works for all fields in SomedayEventForm
- [ ] Works for all fields in EventForm
- [ ] Works for mac and windows

### Steps to Reproduce

1. Create an event (Someday or a regular timed event)
2. Click event to open the form
3. TAB or click to change the element's focus
4. Type `META` + `ENTER`
5. Notice: It'll only submit if the focus is on the title or description elements

### Possible Solution

When implementing your fix, take care to keep the code DRY and to not introduce performance or dependency issues. For example, don't register a new event listener for each element of the form or rely on passing props around everywhere, both of which would result in lots of re-rendering

### Context

This inconsistent behavior leaves users confused why the shortcut doesn't work in all scenarios and makes them less likely to learn or use the shortcuts.

sarans-h:
fixes #458

### DIFF
--- a/packages/web/src/common/types/web.event.types.ts
+++ b/packages/web/src/common/types/web.event.types.ts
@@ -20,6 +20,8 @@ export const SomedayEventSchema = CoreEventSchema.extend({
   isSomeday: z.literal(true),
 });
 
+export type Schema_SomedayEvent = z.infer<typeof SomedayEventSchema>;
+
 export interface Schema_GridEvent extends Schema_Event {
   hasFlipped?: boolean;
   isOpen?: boolean;
@@ -32,12 +34,6 @@ export interface Schema_GridEvent extends Schema_Event {
     initialX: number | null;
     initialY: number | null;
   };
-}
-
-export interface Schema_SomedayGridEvent extends Schema_Event {
-  _id: string;
-  order: number;
-  isSomeday: true;
 }
 
 export interface Schema_OptimisticEvent extends Schema_Event {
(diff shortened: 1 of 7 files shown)


packages/web/src/views/Forms/EventForm/EventForm.tsx
```diff
@@ -186,6 +186,20 @@ export const EventForm: React.FC<FormProps> = ({
...
+        e.preventDefault();
+        onSubmitForm();
+      }
+    };
+    document.addEventListener("keydown", handleMetaEnter);
+    return () => {
+      document.removeEventListener("keydown", handleMetaEnter);
+    };
+  }, [isFormOpen, onSubmitForm]);
+
```
tyler-dane: Adding another `useEffect` just for this use case isn't ideal.

Why not use react hot keys like we're already doing in this file?

sarans-h: done

sarans-h:
@that-one-arab  i forgot to add it for SomedayEventForm but now i added it for it please have a look. 
Thank you

Issue #488, opened 2025-06-06
API logic to save state of sidebar events that were DnD'ed from section to section


that-one-arab:
### Feature Description

Followup to https://github.com/SwitchbackTech/compass/issues/468

Right now, the drag-and-drop functionality is ephemeral. State is not actually saved in the backend.

We need to do the following 2 things:
- Devise a strategy to handle updating start-end dates for sidebar events that were moved from section to section
- Implement API logic to handle saving the updated events based on the devised strategy

PR #503, opened 2025-06-10
🐛 hotfix: skip someday event validation

### LINKED ISSUE #502: Existing someday events fail runtime validation
### Where did this happen?

### Expected Behavior

If a user has existing someday events before the recent code changes in #490, the events should still render in the UI without a runtime error.

### Current Behavior

When loading the web app on `main`, this error appears..

Image

### Steps to Reproduce

1. If you don't already have a someday event without an `order` value, create one in the DB manually
2. Run the web server
3. Go to the web server on localhost
4. Notice the error

### Possible Solution

1. Run a DB migration to populate existing someday events with an order
2. Add a fallback to the UI

### Context

Likely caused by #490

tyler-dane:
Closes #502 by skipping the validation that was causing a runtime error.

### DIFF
--- a/packages/web/src/common/utils/someday.util.ts
+++ b/packages/web/src/common/utils/someday.util.ts
@@ -7,7 +7,6 @@ import {
   Schema_SomedayEventsColumn,
   Schema_SomedayGridEvent,
 } from "@web/common/types/web.event.types";
-import { validateSomedayEvents } from "@web/common/validators/someday.event.validator";
 
 export const getSomedayEventCategory = (
   event: Schema_Event,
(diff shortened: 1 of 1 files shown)


tyler-dane:
@that-one-arab

that-one-arab:
@tyler-dane noted. I also took a look at #502 , if you can fix it in the backend side that would be great, otherwise I could do something like add the `order` field at the `GetSomedayEventsApi` level before feeding it to the rest of the app

tyler-dane:
> @tyler-dane noted. I also took a look at #502 , if you can fix it in the backend side that would be great, otherwise I could do something like add the `order` field at the `GetSomedayEventsApi` level before feeding it to the rest of the app

Do we urgently need to have `order` on all someday events? Is that a prerequisite to finishing #488? If not, I'd rather or hold off altogether

that-one-arab:
@tyler-dane no not really, lets hold it off, but if we don't do it soon enough I'll probably implement a frontend patch to manually add order field to someday events that don't have it

PR #514, opened 2025-06-12
✨ Feat: Add seed command to create sample events in bulk

### LINKED ISSUE #507: Prompt the user for confirmation before running `seed` command
### Use-Case
Since this command will add lots of events to their Compass calendar and Google Calendar, we want users to be sure they know what they're getting into beforehand.

### Implementation
When the user runs the CLI command, we should print a warning in the terminal that explains how this command will
- creates lots of events in their Compass calendar database
- create lots of events in their `primary` Google calendar

The warning should suggest that this is best done on test account rather than a personal account

The prompt should then ask if they want to continue.

pyoneerC:
Closes #507

Changes:
- Added warning message about creation of multiple events in Compass and Google Calendar
- Added recommendation to use test account instead of personal account
- Implemented user confirmation prompt before proceeding
- Added --force flag to skip confirmation
- Added unit tests
- Updated documentation

### DIFF
--- a/packages/scripts/src/commands/seed.ts
+++ b/packages/scripts/src/commands/seed.ts
@@ -1,0 +1,72 @@
+import { _confirm, log } from "@scripts/common/cli.utils";
+import mongoService from "@backend/common/services/mongo.service";
+import eventService from "@backend/event/services/event.service";
+
+const WARNING_MESSAGE = `⚠️ WARNING ⚠️
+
+This operation will:
+1. Create multiple events in your Compass calendar
+2. Create multiple events in your main Google Calendar
+
+It is recommended to use a test account instead of your personal account.
+
+Do you want to continue?`;
+
+export const startSeedFlow = async (force = false) => {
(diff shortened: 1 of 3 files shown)


that-one-arab:
@pyoneerC Thanks for the PR! But I noticed I get this error when running the command

image


Is there a perquisite to this that I should do before running?

that-one-arab (changes requested):
comment above

packages/scripts/src/commands/seed.ts
```diff
@@ -0,0 +1,72 @@
...
+          frequency: "biweekly",
+          interval: 2,
+        },
+      },
+    ];
+
+    for (const eventData of sampleEvents) {
+      await eventService.createEvent(eventData);
+      log.info(`✓ Event created: ${eventData.title}`);
+    }
```
tyler-dane, line 63: For anyone else referencing this PR: 

Let's use `eventService:createMany()` instead of manually looping like this.

PR #520, opened 2025-06-14
fix(focus-event-form-title)

### LINKED ISSUE #283: Event title is not focused when triggered from command palette
### Where did this happen?

### Expected Behavior

Given that the user triggers the action to create a new timed event from the command palette
Then the event form should open for the new event draft
And the focus should be on the event title

### Current Behavior

The title is not focused, and this appears in devtools:

```
Uncaught RangeError: Maximum call stack size exceeded.
    at Object.invokeGuardedCallbackDev (react-dom.development.js:4213:1)
    at invokeGuardedCallback (react-dom.development.js:4277:1)
    at invokeGuardedCallbackAndCatchFirstError (react-dom.development.js:4291:1)
    at executeDispatch (react-dom.development.js:9041:1)
    at processDispatchQueueItemsInOrder (react-dom.development.js:9073:1)
    at processDispatchQueue (react-dom.development.js:9086:1)
    at dispatchEventsForPlugins (react-dom.development.js:9097:1)
    at react-dom.development.js:9288:1
    at batchedUpdates (react-dom.development.js:3985:1)
    at dispatchEventForPluginEventSystem (react-dom.development.js:9287:1)
```

Image

### Steps to Reproduce

Open cmd palette and click or press enter upon the `Create Event [c]` option

### Possible Solution

Upgrade headlessui version

Disable auto focus in cmdk

Wrap in `FocusableGroup` (headless ui option, which is a library that react-cmdk relies on)

Use a div wrapper

### Context

Putting the focus on the title is important to help the user not have to move their mouse and click, which is the whole point of using the command palette.

It is also important to make this timed draft consistent with the other sidebar drafts

victor-enogwe:
# What does this PR do?

This PR fixes the bug "Event title is not focused when triggered from command palette"

## Description

Closes #283 

## Manual Test

https://github.com/user-attachments/assets/374c95f6-7214-4218-81a9-4fd1105f7c10

### DIFF
--- a/packages/web/src/common/types/web.event.types.ts
+++ b/packages/web/src/common/types/web.event.types.ts
@@ -16,9 +16,11 @@ export const GridEventSchema = CoreEventSchema.extend({
 
 export const SomedayEventSchema = CoreEventSchema.extend({
   _id: z.string(),
-  order: z.number(),
+  order: z.number().optional(),
   isSomeday: z.literal(true),
 });
+
+export type Schema_SomedayEvent = z.infer<typeof SomedayEventSchema>;
 
 export interface Schema_GridEvent extends Schema_Event {
   hasFlipped?: boolean;
@@ -32,12 +34,6 @@ export interface Schema_GridEvent extends Schema_Event {
     initialX: number | null;
     initialY: number | null;
   };
-}
-
-export interface Schema_SomedayGridEvent extends Schema_Event {
-  _id: string;
-  order: number;
-  isSomeday: true;
 }
 
 export interface Schema_OptimisticEvent extends Schema_Event {
--- a/packages/web/src/common/utils/__tests__/someday.util.test.ts
+++ b/packages/web/src/common/utils/__tests__/someday.util.test.ts
@@ -5,9 +5,11 @@ import isSameOrBefore from "dayjs/plugin/isSameOrBefore";
 import timezone from "dayjs/plugin/timezone";
 import utc from "dayjs/plugin/utc";
 import { Origin, Priorities } from "@core/constants/core.constants";
+import { Schema_Event } from "@core/types/event.types";
 import { COLUMN_MONTH, COLUMN_WEEK } from "@web/common/constants/web.constants";
-import { Schema_SomedayGridEvent } from "@web/common/types/web.event.types";
+import { Schema_SomedayEvent } from "@web/common/types/web.event.types";
 import { categorizeSomedayEvents } from "../someday.util";
+import { setSomedayEventsOrder } from "../someday.util";
 
 dayjs.extend(utc);
 dayjs.extend(timezone);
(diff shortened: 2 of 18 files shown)


packages/web/src/common/utils/onEventTargetVisibility.ts
```diff
@@ -0,0 +1,25 @@
...
+  ) => {
+    const observer = new IntersectionObserver(([entry]) => {
+      if (entry.isIntersecting !== visible) return;
+
+      observer.disconnect();
+      callback();
+    });
+
+    observer.observe(event.currentTarget);
+  };
```
that-one-arab: - The file is a little misplaced, as files in this directory follow the `[NAME].util.ts` convention

that-one-arab: Could you expand upon why we needed to implement this logic to get the input focus to work?

that-one-arab (changes requested):
Thanks for the PR!

I tested your solution and it works, plus tests are passing as well!

I just need you to clarify a question I left you in the code review when you can please. This will help us gauge the quality of the fix in this PR

packages/web/src/common/utils/onEventTargetVisibility.ts
victor-enogwe: Sure!

The `CommandPalette` has an input field which is focused once you hit `Meta+k`. It also closes once you click its menu item.

Once you click the `Create Event` button within the command palette, it begins to hide(`close`) the `CommandPalette` starting with appending an `aria-hidden="true"` attribute to the  `CommandPalette` root element.

The bug surfaces because the action within the `onClick` handler completes before the palette closes(you get a race condition here).

Opening the `EventForm` moves the focused element to the `title input` field within it. Since the `CmdPalette` visibility state has not transitioned before the attempt to move focus, the `aria-hidden="true"` attribute is applied to the `CmdPalette` and its input is still focused , this `chromium` browser rule shown in the image below, will prevent the focus transition.
Screenshot 2025-06-16 at 3 26 08 AM

To fix this, a visibility observer is initialized within the `onClick` handler. It monitors the command palette visibility for changes(until it is hidden). Once the palette is hidden, the desired action for the button(`onEventTargetVisibility callback`) is triggered and the visibility observer is disconnected.

victor-enogwe: @that-one-arab I've update the filename, let me know if it meets the convention.

that-one-arab: @victor-enogwe Thats a great solution! Thanks for your work in debugging this!

If it won't trouble you, could you please link this github comment as a JS comment above the function?

Something like this
```
// See this for additional context: https://github.com/SwitchbackTech/compass/pull/520#discussion_r2148965613
export const onEventTargetVisibility =
... rest of code...
```

This would help other devs know the purpose behind this.

After that, the PR is good to go!

that-one-arab (changes requested):
https://github.com/SwitchbackTech/compass/pull/520#discussion_r2150375132

packages/web/src/common/utils/onEventTargetVisibility.ts
victor-enogwe: Hi @that-one-arab I've pushed the requested comment changes. Please take a look.

Issue #508, opened 2025-06-11
Authenticate user before creating events during the `seed` command


tyler-dane:
### Feature Description

As a CLI command, I want to be able to authenticate a user to Compass' backend (via Google/Supertokens), so that I can pass the Compass identify verification middleware and access its APIs.

### Use Case

The `seed` command will create Compass and Google events. Using the Compass API to create the events is the easiest way to do this, as all of the business logic required to map events to their correct schemas is already handled there.

In order for this CLI to make calls to the Compass API, however, you'll need to authenticate with Supertokens and Google calendar. These security measures ensure that a client can only update their events.

### Scope
This issue can be considered done once the command successfully creates a session and prints the user's id to the terminal.

### Additional Context

**How Session Authentication Works**

TL;DR:
1. Create a session to get an access token
2. Use the access token in subsequent requests

----

To see what happens when you try to make an unauthorized call to the Compass API, run:

```sh
curl --location 'http://localhost:3000/api/event' \
--header 'Content-Type: application/json' \
--data '{
    "gEventId": "get-this-from-gcal",
    "user": "your-user-id",
    "description": "this is an event that was first created in gcal",
    "endDate": "2025-06-11T16:15:00-05:00",
    "isAllDay": false,
    "origin": "google",
    "priority": "work",
    "startDate": "2025-06-11T15:45:00-05:00",
    "title": "Timed Event"
}'
```

You'll get a 401 response with the body:

```json
{
    "message": "unauthorised"
}
```

This error occurs, because we're verifying that a user session exists on the server (`event.routes.config.ts:verifySession()`) before letting the request make it to the controller.

Run this command to create a user session using cookie-based auth on the server, replacing the `email` value with your account:

```sh
curl -i -c --location 'http://localhost:3000/api/auth/session' \
--header 'rid: session' \
--header 'st-auth-mode: cookie' \
--header 'Content-Type: application/json' \--data-raw '{
    "email": "your-gmail-account@gmail.com"
}'
```

You'll get a response like this:

```text
HTTP/1.1 200 OK
Access-Control-Allow-Origin: http://localhost:9080
Vary: Origin
Access-Control-Allow-Credentials: true
Content-Security-Policy: default-src 'self';base-uri 'self';font-src 'self' https: data:;form-action 'self';frame-ancestors 'self';img-src 'self' data:;object-src 'none';script-src 'self';script-src-attr 'none';style-src 'self' https: 'unsafe-inline';upgrade-insecure-requests
[...]

PR #541, opened 2025-06-20
✨ Feat: Add user authentication for seed command

claudiahdz:
# Summary
This PR adds user authentication on the CLI in order to call the Compass API and create events for the `seed` command. 

Run using: `yarn cli seed -u <userEmail | userId>`

# Changes:
- Created a `cli.auth.js` file that manages the creation of a user session and returns the access token
- Updated `getAPIBaseUrl` and other associated variables and functions to work for local environment
- Added a single test event to the seed command that calls the new `createEvent` function that calls the API

## Example
Screenshot 2025-06-20 at 11 48 09 a m

# 
Closes https://github.com/SwitchbackTech/compass/issues/508


packages/scripts/src/commands/seed.ts
```diff
@@ -1,10 +1,34 @@
...
+import { FORMAT } from "@core/util/date/date.util";
 import mongoService from "@backend/common/services/mongo.service";
 import { findCompassUserBy } from "@backend/user/queries/user.queries";
 
 const { prompt } = pkg;
 
+async function createEvent(events: Schema_Event[], email: string) {
+  try {
+    const accessToken = await createSession(email);
+    const baseUrl = await getApiBaseUrl("local");
```
claudiahdz: Seeds events only on local.

packages/scripts/src/common/cli.auth.ts
tyler-dane, line 1: This has a lot of overlap with `auth.controller:createSession`. Can you move what's needed into the auth controller/service? I'd rather have that auth functionality contained in the auth portion of the code than in a CLI util file. Doing so will make it easier to reuse auth functionality (for example once we have a mobile app).

Feel free to adjust the Auth Controller if needed. It's currently not hooked up to any other functionality in the app, so there's no risk of it breaking downstream features. However, try to rely on the Supertokens API to do most of the heavy lifting. For example, I think you can use `Session.createNewSession(req, res, "public", sUserId)`  to avoid manually setting cookies.

Here's an example of how our existing `/session` endpoint works

Screenshot 2025-06-20 at 6 10 13 AM



(Apologies for not mentioning that controller method initially. TBH I forgot it was there)

packages/scripts/src/commands/seed.ts
```diff
@@ -22,11 +46,42 @@ async function seedEvents(userInput: string) {
...
+          .minute(0)
+          .second(0)
+          .format(FORMAT.RFC3339_OFFSET.value),
+        endDate: dayjs()
+          .hour(11)
+          .minute(0)
+          .second(0)
+          .format(FORMAT.RFC3339_OFFSET.value),
+      },
+    ];
```
tyler-dane: Instead of creating an event from scratch, please use the factory methods from `ccal.event.factory` and override any properties as needed. This'll make it easier to keep our seed data updated as the schema changes.

packages/scripts/src/commands/seed.ts
```diff
@@ -22,11 +46,42 @@ async function seedEvents(userInput: string) {
...
+    await createEvent(event, user.email);
 
-    log.success(`Successfully created events for user: ${userInput}`);
+    log.success(
+      `Successfully created events for user: ${user.email} with id: ${userId}`,
+    );
   } catch (error) {
```
tyler-dane, line 71: Ran into an error when trying this command. Same error occurred when attempting to use the userid instead of the email.

err

packages/scripts/src/commands/seed.ts
```diff
@@ -1,10 +1,34 @@
...
+        "Content-Type": "application/json",
+        Cookie: `sAccessToken=${accessToken}`,
+      },
+    });
+    return response.data;
+  } catch (error) {
+    console.error("Failed to create events:", error);
+    throw error;
+  }
+}
```
tyler-dane, line 22: Instead of making the seed command worry about headers, creating events, and error handling itself, can we call our `event.controller:create` instead?

packages/scripts/src/common/cli.constants.ts
```diff
@@ -15,6 +15,8 @@ export const CATEGORY_VM = {
 };
 
 export const CLI_ENV = {
+  LOCAL_DOMAIN:
+    process.env["LOCAL_DOMAIN"] || `localhost:${process.env["PORT"]}`,
```
tyler-dane: please add `LOCAL_DOMAIN` to the `.env.example` file so newcomers are aware of this option

packages/scripts/src/commands/seed.ts
claudiahdz: @tyler-dane Q: You mean the `event.service:create`? If that's the case, I'm a bit confused as to why we needed the authentication part.

tyler-dane: Yes, `event.controller` calls `event.service:create` (or `event.service:createMany`, depending on the request).

The reason we need authentication is because in order to get to the controller, we need to get past the middleware that requires an authenticated session:

```ts
//event.routes.config.ts

  configureRoutes(): express.Application {
    this.app
      .route(`/api/event`)
      .all(verifySession())
      //@ts-ignore
      .get(eventController.readAll)
      //@ts-ignore
      .post(eventController.create);
```

tyler-dane: We need to ensure that one user cannot create events for another user's account

claudiahdz: I simplified the controller call!

packages/scripts/src/commands/seed.ts
```diff
@@ -22,11 +37,32 @@ async function seedEvents(userInput: string) {
...
+
+    // Test Event
+    const eventOverrides = {
+      user: userId,
+      isAllDay: false,
+      isSomeday: false,
+      startDate: dayjs().hour(10).minute(0).second(0).toISOString(),
+      endDate: dayjs().hour(11).minute(0).second(0).toISOString(),
+    };
+    const event = createMockStandaloneEvent(eventOverrides);
```
claudiahdz, line 61: Using the event factory now.

packages/scripts/src/commands/seed.ts
claudiahdz: Marking as resolved, only because the auth approached changed.

packages/scripts/src/common/cli.auth.ts
claudiahdz: I introduced a new function called `createSessionForUser` on the auth service. This functions calls `Session.createNewSessionWithoutRequestResponse` and retrieves the access token using the secure `getAccessToken()` from supertokens.

packages/scripts/src/commands/seed.ts
claudiahdz: I'm using now the factory!

claudiahdz:
@tyler-dane Hi! So I updated the implementation to take care of the feedback you left. Let me know what you think!

packages/backend/src/auth/services/compass.auth.service.ts
```diff
@@ -28,6 +32,45 @@ class CompassAuthService {
...
+      userId = cUserId;
+    } else if (email) {
+      const user = await findCompassUserBy("email", email);
+      if (!user) {
+        throw new Error("User doesn't exist");
+      }
+      userId = user._id.toString();
+    } else {
+      throw new Error("Unable to determine user ID");
+    }
```
tyler-dane: I'm concerned about the security implications of allowing someone to check for a user based on an email string from the CLI. (This was why I merged #545 recently)

Please only rely on the user id to find the user

packages/scripts/src/commands/seed.ts
tyler-dane: Thank you. 
However, there's now a lot of duplication across the auth controller and service. Can you simplify the controller, so that it simply calls the service?

Screenshot 2025-06-23 at 7 37 43 PM

packages/scripts/src/commands/seed.ts
tyler-dane: Please type `eventOverrides`, so that we remember to update it when the schema changes.

packages/scripts/src/common/cli.constants.ts
tyler-dane: I see you updated the `.env.example` to use `http://localhost:3000`. That makes me that that the currently fallback of `localhost:${process.env["PORT"]}` might be incorrect? Shouldn't it be `http://localhost:3000`?

claudiahdz: You're right, I've made the proper changes!

packages/scripts/src/commands/seed.ts
claudiahdz: Now controller calls service.

packages/scripts/src/commands/seed.ts
claudiahdz: I've making changes on the implementation for creating multiple events in the following PR. I'm adding a template system to easily add events. Is it okay to revisit in the next PR?

packages/backend/src/auth/services/compass.auth.service.ts
claudiahdz: I removed email from the service!

packages/scripts/src/commands/seed.ts
claudiahdz: For reference: https://github.com/SwitchbackTech/compass/pull/562

packages/backend/src/auth/services/compass.auth.service.ts
tyler-dane: I was referring to allowing the user to enter their email as a CLI input. 
For the sake of time, I implemented that fix here [9a3996b](https://github.com/SwitchbackTech/compass/pull/541/commits/9a3996b4aa8dbbaf98dd918e9638120288aa5765)

packages/scripts/src/commands/seed.ts
tyler-dane: Sounds good! Thanks for including the link to the follow-up issue

PR #547, opened 2025-06-21
✨ feat: add platform-specific keyboard shortcuts for cmd palette tooltip

fogboots:
- Updated command palette tooltip to show Cmd+K on macOS and Win+K on Windows
- Replace hardcoded 'CMD + K' text with dynamic icons based on user's OS
- Use Command icon for macOS and WindowsLogo icon for Windows users from Phosphor Icons
- Consistent with existing platform-specific shortcuts in the app

I have yet to test on windows.


that-one-arab:
@fogboots Thanks for the PR! I don't think we need to wait on you to test this we can count on `getDesktopOS` function being dependable and returning the correct OS. We will revisit if we prove that is not the case.

PR #558, opened 2025-06-24
Feat optimize event import performance

### LINKED ISSUE #544: Optimize event import performance
### Feature Description

Update backend code to reduce how long it takes for calendar events to finish being imported into Compass.

### Use Case

This'll improve the onboarding UX by lettings users see and use their existing calendar events quicker.

### Additional Context

This should be worked on after #543 is finished. 

### Implementation Notes
Here are some ideas on how to improve performance:

Instead of a general `fetchAndProcessEventsPageByPage` call with `singleEvents: true` (which fetches all instances and re-fetches regular events), iterate over the `recurringEventIds` collected in Pass 1.

For each `recurringEventId`, make a specific API call to the Google Calendar API to fetch only the instances for that particular master event (e.g., using `calendar.events.instances({ calendarId, eventId: masterEventId, ... }))`.

These individual API calls for each master event's instances can be executed in parallel (e.g., using `Promise.allSettled` for robustness against individual failures).

The processing logic (`shouldProcessDuringPass2`) would then handle these fetched instances.

After all parallel instance fetching and processing is complete, make a separate, lightweight `events.list` call (e.g., with maxResults: 1 and singleEvents: true) to obtain the `nextSyncToken` for the calendar.

victor-enogwe:
# What does this PR do?

This PR optimizes the backend calendar import/full sync process to reduce how long it takes for calendar events to finish being imported into Compass.

## Description

Closes #544 

## Manual Test

- Create a test `GCAL` account with more than 50,000 events.
- Signup with the test account
- Check that the calendar import takes less than (25 secs)


victor-enogwe:
@tyler-dane please take a look.
It's still a work in progress and requires feedback.

- I've taken into account imports being stalled or preempted, the `baseEvents` import rely on the `nextPageToken` and `nextSyncToken`. The `nexPageToken` will be stateful like the `nextSyncToken` and can be used to resume importing from last `page fetched` instead of starting the fetch from the first page - **This is still WIP**.
- In future iterations, the UI may benefit from a limited import(the current week's event) during signup, so the user can see events on their dashboards - assuming a non-empty calendar.

Questions:
- Is there a need to limit the recurring events fetched as it is done here [ecur.manager.ts](https://github.com/SwitchbackTech/compass/blob/0224528580b253e75caa111c6bd00273eb3d0329/packages/backend/src/event/services/recur/recur.manager.ts#L183)?
- What would be considered as ideal timing(in seconds) for importing e.g. `50,000` events?

victor-enogwe:
@tyler-dane I'm not sure how refecthing the next sync token via the `events.list` gcal API will work using `singleEvents: true`. The call would start at the begining of the list, since we do not have access to a `nextPageToken`. For large calendars with events greater than the per-page limit, It would also not return the `nextSyncToken` and would require iterating till the end of the list till we get this token. Do you have ideas on how we can get the `nextSyncToken` at the end without iterating through the end of the list?

I am currently testing to see if the `nextSyncToken` from the  call using `singleEvents: false`  token can be used as 
can return accurately in  subsequent list calls using `singleEvents: true`.

packages/backend/src/common/services/gcal/gcal.service.ts
```diff
@@ -46,6 +69,98 @@ class GCalService {
...
+        typeof nextPageToken === "string" && nextPageToken.length > 0;
+
+      yield { nextPageToken, nextSyncToken, items };
+    } while (hasNextPage);
+  }
+
+  /**
+   * getAllEvents
+   * generator function to list all google calendar events
+   * @param param0
```
tyler-dane: What does `@param param0` mean? Please update for clarity

packages/backend/src/sync/services/import/sync.import.full.test.ts
```diff
@@ -40,54 +40,63 @@ describe("SyncImport: Full", () => {
...
       await syncImport.importAllEvents(setup.userId, "test-calendar");
 
     const currentEventsInDb = await mongoService.event.find().toArray();
 
-    expect(totalProcessed).toBe(5); // base + 2 instances + regular + cancelled
-    expect(totalChanged).toBe(4); // base + 2 instances + regular - cancelled
-    expect(currentEventsInDb).toHaveLength(4); // base + 2 instances + regular - cancelled
+    expect(totalProcessed).toBe(6); // base + 3 instances + regular + cancelled
+    expect(totalChanged).toBe(5); // base + 3 instances + regular
+    expect(currentEventsInDb).toHaveLength(5); // base + 3 instances + regular
```
tyler-dane, line 70: Why did these numbers change? Is this related to the changes in `gcal.factory.ts`?

tyler-dane:
> * What would be considered as ideal timing(in seconds) for importing e.g. `50,000` events?

I don't have a good answer for that. But in general, I'd like to keep the time that a user has to wait before being able to access Compass under 10 seconds

packages/backend/src/sync/services/import/sync.import.full.test.ts
victor-enogwe: Yes, the switch to  the [`event.instances`](https://github.com/SwitchbackTech/compass/blob/8a397eea3b5a4ef2d2099b79668163b928567143/packages/backend/src/__tests__/mocks.gcal/factories/gcal.factory.ts#L114) API.

packages/backend/src/common/services/gcal/gcal.service.ts
victor-enogwe: Editor auto comment - fixed

packages/backend/src/sync/services/import/sync.import.ts
```diff
@@ -276,70 +272,154 @@ export class SyncImport {
     return response.data;
   }
 
+  async importEventInstances(
```
tyler-dane, line 250: This is a bit hard to follow. Please split this up into at least 2 separate functions: one to assemble the correct `processedEvents` and the other to create them in the DB.

packages/backend/src/sync/services/import/sync.import.ts
```diff
@@ -276,70 +272,154 @@ export class SyncImport {
...
+
   /**
-   * Imports ALL events for a calendar (Full Sync).
+   * importAllEvents
+   * Import ALL events for a calendar (Full Sync).
+   * @param userId
+   * @param calendarId
+   * @param perPage
+   * @returns
    */
```
tyler-dane, line 300: Please clean this up. If we're gonna include doc string params, they need to add clarity. This is simply repeating the function name and signature.

packages/backend/src/common/services/gcal/gcal.service.ts
```diff
@@ -1,12 +1,31 @@
...
+import type { Params_WatchEvents } from "@core/types/sync.types";
 import { GCAL_PRIMARY } from "@backend/common/constants/backend.constants";
 import { ENV } from "@backend/common/constants/env.constants";
 import { error } from "@backend/common/errors/handlers/error.handler";
 import { GcalError } from "@backend/common/errors/integration/gcal/gcal.errors";
 
 class GCalService {
+  private validateGCalResponse<T>(
+    response: GaxiosResponse<T>,
+    message = "Failed to fetch GCal items",
```
tyler-dane: Since this is a generic validator that only relies on the status, let's update the message to it to apply to other gcal operations (not just fetching)

packages/backend/src/sync/services/import/all/import.all.util.test.ts
```diff
@@ -1,86 +0,0 @@
...
-  it("assigns Mongo ObjectIds to events", () => {
-    const events = getEventsWithoutIds() as Event_Core[];
-    assignIds(events);
-
-    // Verify all events have ObjectIds
-    for (const event of events) {
-      expect(event?._id).toBeDefined();
-      expect(event._id).toBeInstanceOf(ObjectId);
-    }
-  });
```
tyler-dane, line 46: Verifying that each of the processed events has a valid ObjectId is still an important part of the import process that we need to test. When you split the import functions (see [this comment](https://github.com/SwitchbackTech/compass/pull/558/files#r2168771510)), please be sure to add a test that validates the id like we used to do here.

packages/backend/src/sync/services/import/sync.import.ts
```diff
@@ -276,70 +272,154 @@ export class SyncImport {
...
+    calendarId: string,
+    baseEvent: gSchema$Event,
+    perPage = 1000,
+  ): Promise<{
+    totalProcessed: number;
+    totalSaved: number;
+    totalInstancesSaved: number;
+  }> {
+    const isRecurringEvent = Array.isArray(baseEvent.recurrence);
+    const baseEventId = new ObjectId();
```
tyler-dane: This is actually a `standalone` event, not a `baseEvent`. Since there are so many event variations, we need to be really careful how we use these terms to avoid confusion. 

Please see these docs to understand the difference and update the other changes in this PR accordingly

https://docs.compasscalendar.com/docs/learn/glossary
https://docs.compasscalendar.com/docs/learn/recurring-events-overview

packages/backend/src/sync/services/import/sync.import.full.test.ts
```diff
@@ -40,54 +40,63 @@ describe("SyncImport: Full", () => {
...
     // Verify we have the base recurring event
-    const baseEvents = currentEventsInDb.filter(
-      (e) => e.recurrence?.rule !== undefined,
+    const baseEvents = currentEventsInDb.filter((e) =>
+      Array.isArray(e.recurrence?.rule),
     );
```
tyler-dane: There are a lot of cases in this PR where we're using `isArray(e.recurrence?.rule)` to determine if an event is recurring or not. Let's create a function so we can encapsulate that logic in one place. You could call it something like `isRecurringGcalEvent` and put it in `recur.gcal.util.ts` (separate from the class).

packages/backend/src/__tests__/mocks.gcal/factories/gcal.factory.ts
```diff
@@ -78,16 +80,21 @@ export const mockGcal = ({
                 pageToken?: string;
                 singleEvents?: boolean;
               }) => {
-                // When singleEvents is false, only return base events
-                if (params.singleEvents === false) {
-                  const baseEvents = events.filter((event) => event.recurrence);
+                // When singleEvents is false, only return base events and non-recurring events
+                if (!params.singleEvents) {
+                  const baseEvents = events.filter(
+                    (event) => event.recurrence || !event.recurringEventId,
```
tyler-dane: An event is not a baseEvent just because it doesn't have `recurringEventId`.
- An event is a base event if it has a recurrence AND does not have a `recurringEventId`

packages/backend/src/common/services/gcal/gcal.service.ts
victor-enogwe: done

packages/backend/src/sync/services/import/sync.import.ts
victor-enogwe: done

packages/backend/src/sync/services/import/sync.import.full.test.ts
victor-enogwe: done. they now use event discriminator functions within the [event.utils.ts](https://github.com/SwitchbackTech/compass/blob/main/packages/core/src/util/event/event.util.ts)  and [gcal.event.utils.ts](https://github.com/SwitchbackTech/compass/blob/main/packages/core/src/util/event/gcal.event.util.ts) files

packages/backend/src/sync/services/import/sync.import.ts
victor-enogwe: I've renamed the variable  `isRecurringEvent` to `isBaseEvent` and now use the checks within the [event.util.ts](https://github.com/SwitchbackTech/compass/blob/main/packages/core/src/util/event/event.util.ts) file.

packages/backend/src/__tests__/mocks.gcal/factories/gcal.factory.ts
victor-enogwe: fixed.

packages/backend/src/sync/services/import/sync.import.ts
victor-enogwe: refactored

packages/backend/src/sync/services/import/all/import.all.util.test.ts
victor-enogwe: tests added to the `sync.import.util.test`

PR #559, opened 2025-06-24
✨ Feat(web): Add move to next/prev week/month shortcuts for someday event

### LINKED ISSUE #526: Add shortcuts to migrate someday event buttons
### Feature Description

As a user, I want to quickly be able to move a someday event forward/backwards via button click or shortcut, so that I can customize my schedule easily as plans change.

----

Followup to https://github.com/SwitchbackTech/compass/issues/525

After the buttons are implemented in the someday event form, we should add the following shortcuts to the buttons:
- `CTRL + META + RIGHT ARROW`: Migrate forward 1 week/month
- `CTRL + META + LEFT ARROW`: Migrate backwards 1 week/month.

These shortcuts are meant to compliment the  `CTRL+META+UP/DOWN` ones for migrating between weeks/month lists (#527 )

### Use Case

This'll save users time by providing better accessibility options, making the program easier and more fun to use.

### Additional Context

_No response_

that-one-arab:
## Description

Closes #526

### DIFF
--- a/packages/web/src/common/utils/web.date.util.ts
+++ b/packages/web/src/common/utils/web.date.util.ts
@@ -375,3 +375,48 @@ export const setEventStartEndDatesToCurrentMonth = (
     endDate: monthEnd.format(),
   };
 };
+
+export const setEventStartEndDates = (
+  to: {
+    direction: "prev" | "next" | "current";
+    duration: "week" | "month";
+  },
+  event: Schema_Event,
+): Schema_Event => {
+  const reference =
+    to.direction === "current" ? dayjs(new Date()) : dayjs(event.startDate);
+
+  let start: Dayjs;
+  let end: Dayjs;
+
+  if (to.duration === "week") {
+    start = reference.startOf("week");
+    end = reference.endOf("week");
+
+    if (to.direction === "prev") {
+      start = start.subtract(1, "week");
+      end = end.subtract(1, "week");
+    } else if (to.direction === "next") {
+      start = start.add(1, "week");
+      end = end.add(1, "week");
+    }
+  } else {
+    // duration is month
+    start = reference.startOf("month");
+    end = reference.endOf("month");
+
+    if (to.direction === "prev") {
+      start = start.subtract(1, "month");
+      end = end.subtract(1, "month");
+    } else if (to.direction === "next") {
+      start = start.add(1, "month");
+      end = end.add(1, "month");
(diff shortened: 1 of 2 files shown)


packages/web/src/common/utils/web.date.util.ts
```diff
@@ -375,3 +375,48 @@ export const setEventStartEndDatesToCurrentMonth = (
...
+
+export const setEventStartEndDates = (
+  to: {
+    direction: "prev" | "next" | "current";
+    duration: "week" | "month";
+  },
+  event: Schema_Event,
+): Schema_Event => {
+  const reference =
+    to.direction === "current" ? dayjs(new Date()) : dayjs(event.startDate);
```
tyler-dane, line 387: Just curious, why use `dayjs(new Date())` instead of `dayjs()`?

packages/web/src/common/utils/web.date.util.ts
```diff
@@ -375,3 +375,48 @@ export const setEventStartEndDatesToCurrentMonth = (
     endDate: monthEnd.format(),
   };
 };
+
+export const setEventStartEndDates = (
```
tyler-dane, line 379: Let's add some unit tests for this, as I'm already seeing some bugs with its implementation

tyler-dane, line 379: Can you rename this so it doesn't start with `set...`. Since this isn't changing any React state, I think we should stay away from `set...` as the first part of a function to avoid confusion

packages/web/src/common/utils/web.date.util.ts
that-one-arab: Uhhhh no particular reason, I guess I was used to it but now after reading your comment I realized that the latter is the better practice, thanks for the headsup

packages/web/src/common/utils/web.date.util.ts
that-one-arab: For sure, but can you share what bugs did you stumble upon?

that-one-arab: Yes makes sense, thanks for pointing this out.

tyler-dane: #560

tyler-dane: #561

PR #563, opened 2025-06-25
Feat render calendar after auth - backend

### LINKED ISSUE #543: Render calendar immediately after authenticating
Restructure the login flow so that the user can use Compass calendar immediately after successfully authenticating, while also importing their existing calendar data in the background.

### Implementation Notes
Extract the event imports from the sign up flow
- Trigger the event import as a separate request, so it does not block UI interactions.  
  - Create a new route for importing events   
- Add a subtle loading indicator to the UI to indicate that we're fetching in the background

Fetch data after import finished 
- After the backend finishes importing all events, emit the `EVENT_CHANGED` websocket notification from the server to the client.
  - See `websocket.server.ts` and `SocketProvider.tsx`
- This will trigger the frontend to refetch the current week's events, getting the user's full calendar data.

### Loading Spinner Design

Image

- Add tooltip that says, "Importing your calendar events in the background"
- If you have time, try to animate the spinner so that it feels dynamic. If not, we can descope that and create a separate issue.
- Spinner should disappear after the background import finishes

victor-enogwe:
# What does this PR do?

This PR restructures the login flow so that the user can use Compass calendar immediately after successfully authenticating, while also importing their existing calendar data in the background.

## Description

Closes #543

## Manual Test

Relies on PR #568


packages/backend/src/__tests__/drivers/base.driver.ts
tyler-dane, line 1: It looks like this and the Sync Controller driver aren't being used anywhere. Can you clarify your plan for these? Would each controller have its own driver? Would other parts of the backend use it (not just controllers)?

packages/backend/src/sync/services/import/sync.import.series.test.ts
```diff
@@ -66,7 +51,7 @@ describe("SyncImport: Series", () => {
     expect(baseEvents).toHaveLength(1);
 
     const instancesInDb = filterExistingInstances(currentEventsInDb);
-    expect(instancesInDb).toHaveLength(expectedInstances);
+    expect(instancesInDb).toHaveLength(3);
```
tyler-dane, line 60: It'll be confusing to newcomers where these numbers are coming from. Is there a way you can make it more clear?

packages/backend/src/__tests__/drivers/base.driver.ts
tyler-dane: Update - I see some usage of them in the tests now, disregard

victor-enogwe:
@tyler-dane this is ready for review too

packages/backend/src/sync/controllers/sync.controller.test.ts
```diff
@@ -0,0 +1,155 @@
...
+      expect(notifySpy).toHaveBeenCalledTimes(6);
+
+      Array(6).forEach((_, index) =>
+        expect(notifySpy).toHaveBeenNthCalledWith(
+          index + 1,
+          EVENT_CHANGED,
+          socketId,
+        ),
+      );
+    });
```
tyler-dane, line 146: A couple issues with this test:

1) It uses a magic number (6), which is difficult to understand 

2) It seems like this test is dependent on the other tests in this suite. When I run it in isolation, it fails with

```

  ● SyncController › importGCal › should notify the frontend to refetch the calendar events on completion

    expect(jest.fn()).toHaveBeenCalledTimes(expected)

    Expected number of calls: 6
    Received number of calls: 1

      142 |
      143 |     it("should notify the frontend to refetch the calendar events on completion", async () => {
    > 144 |       expect(notifySpy).toHaveBeenCalledTimes(6);
          |                         ^
      145 |
      146 |       Array(6).forEach((_, index) =>
      147 |         expect(notifySpy).toHaveBeenNthCalledWith(

      at Object.toHaveBeenCalledTimes (packages/backend/src/sync/controllers/sync.controller.test.ts:144:25)

Test Suites: 1 failed, 1 total
Tests:       1 failed, 5 skipped, 6 total
Snapshots:   0 total
```

packages/backend/src/sync/services/sync/__tests__/gcal.sync.processor.delete.test.ts
```diff
@@ -122,11 +114,21 @@ describe("GcalSyncProcessor: DELETE", () => {
...
-    expect(remainingEvents[0]?._id.toString()).toBe(compassBaseId);
+
+    expect(remainingEvents).toEqual(
+      expect.arrayContaining([
+        expect.objectContaining({
+          _id: expect.any(String),
+        }),
+      ]),
+    );
+
```
tyler-dane, line 122: This was originally validating that only the instance was deleted.

It did that by asserting that the remaining events included the compass base. 

This is currently testing that all the remaining events have and _id, which is much less useful. Please refactor to validatie the original intent of this test.

victor-enogwe: Thanks for catching this.

packages/backend/src/sync/controllers/sync.controller.test.ts
victor-enogwe: fixed

packages/backend/src/sync/services/sync/__tests__/gcal.sync.processor.delete.test.ts
victor-enogwe: fixed

victor-enogwe:
Hi @tyler-dane I've resolved your comments. Take a look.

PR #568, opened 2025-06-27
Feat render calendar after auth - ui

### LINKED ISSUE #543: Render calendar immediately after authenticating

victor-enogwe:
# What does this PR do?

This PR adds a subtle loading indicator to the UI to indicate that we're importing the user's GCAL in the background

## Description

Closes #543 - UI stories

## Manual Test

- Sign up as a new user
- Once you're logged in, confirm that a loading spinning icon appears in the sidebar bottom row
- Hover over the loading spinning icon and confirm that it has a tooltip with the text - "Importing your calendar events in the background"
- Confirm that the spinning loading icon disappears within 30 secs - indicating that the import has completed.

https://github.com/user-attachments/assets/97742e0e-6472-4b4b-938d-78f8cbe26f36


victor-enogwe:
@tyler-dane I'm triggering the sync via a a new backend socket event `IMPORT_GCAL` instead of relying on `UI` state.
This event can be called after the user signs up.

We can:
- Deterministically send this socket event  `IMPORT_GCAL` after we're sure that the user's socket has connected. This means we need to keep state(`firstLogin` - boolean) on the backend, so this import will not be arbitrarily repeated on successive connections.
- Non-Deterministically send this socket event  `IMPORT_GCAL` after user sign up, via cron, queue, or staggered event - We'll need to do it within the timeframe we think the user's browser will be open and connected to the socket.
- Allow the user a manual option to trigger this import once they login(a popup or snack notification can suffice).

Note that this will be in a separate issue and PR.

What do you think?

packages/web/src/ducks/events/slices/sync.slice.ts
```diff
@@ -8,15 +9,20 @@ type Payload_TriggerFetch = {
 interface State_Sync {
   isFetchNeeded: boolean;
   reason: null | Sync_AsyncStateContextReason;
+  importGCal?: unknown;
```
tyler-dane: Can we add stricter typing for `importGcal`?

victor-enogwe: I cleaned this up.

victor-enogwe:
Hi @tyler-dane I'm adding that feature

> @tyler-dane I'm triggering the sync via a a new backend socket event `IMPORT_GCAL` instead of relying on `UI` state. This event can be called after the user signs up.
> 
> We can:
> 
> * Deterministically send this socket event  `IMPORT_GCAL` after we're sure that the user's socket has connected. This means we need to keep state(`firstLogin` - boolean) on the backend, so this import will not be arbitrarily repeated on successive connections.
> * Non-Deterministically send this socket event  `IMPORT_GCAL` after user sign up, via cron, queue, or staggered event - We'll need to do it within the timeframe we think the user's browser will be open and connected to the socket.
> * Allow the user a manual option to trigger this import once they login(a popup or snack notification can suffice).
> 
> Note that this will be in a separate issue and PR.
> 
> What do you think?

It's related to this comment. The events fetch needs to be triggered. I'll add a temp trigger trigger for now and create another issue to address it fully

packages/backend/src/auth/controllers/auth.controller.ts
```diff
@@ -181,6 +182,13 @@ class AuthController {
   signup = async (gUser: TokenPayload, gRefreshToken: string) => {
     const userId = await userService.initUserData(gUser, gRefreshToken);
 
+    // @TODO - create a permanent notification trigger for gCal import
+    // temporary trigger for gCal import
+    // we will not await this promise
+    new Promise((resolve) => {
+      const timeout = setTimeout(() => resolve(clearTimeout(timeout)), 5000);
+    }).then(() => webSocketServer.handleImportGCalStart(userId));
+
```
victor-enogwe, line 191: @tyler-dane This is a temprary trigger that will start the import 5 seconds after signup(so that the websocket connection is presumably ready). It will notify the UI to call the importGCal API endpoint.

Issue #569, opened 2025-06-27
Someday event is in the wrong list after using `m` shortcut


tyler-dane:
### Where did this happen?

Local code with an up-to-date main branch

### Expected Behavior

When using the `m` shortcut to create a someday event, the event should be added to the month list.

### Current Behavior

When using the `m` shortcut to create a someday event, the event is added to the week list.

### Steps to Reproduce

Follow exact steps shown here:

https://github.com/user-attachments/assets/a2be4d44-5423-4418-b37b-1e557f1a2b5d

### Possible Solution

We might need to consider the current week in view before blinding assigning the `startDate` and `endDate`s. 
Or perhaps the issue is afterwards after we're categorizing the events in the selector.

If we're in the first week of a month and a user tries to add a month event, we need a way to ensure the event is stored in its intended location.

### Context

Possibly related to #567 (#560)

Ensure this use-case is tested after fixed to prevent regressions.

PR #572, opened 2025-06-29
🐛 Fix: Someday event in wrong location after creating with 'm' shortcut

that-one-arab:
## Description
Closes https://github.com/SwitchbackTech/compass/issues/569
This was due to use not computing the start-end date correctly when the someday event category is month

## Videos

https://github.com/user-attachments/assets/7cdb5db2-c679-4572-b2bd-2a419df19496

### DIFF
--- a/packages/web/src/common/utils/draft/someday.draft.util.test.ts
+++ b/packages/web/src/common/utils/draft/someday.draft.util.test.ts
@@ -1,0 +1,106 @@
+import dayjs from "dayjs";
+import { Categories_Event } from "@core/types/event.types";
+import { draftSlice } from "@web/ducks/events/slices/draft.slice";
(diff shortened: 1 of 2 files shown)


packages/web/src/common/utils/draft/someday.draft.util.test.ts
```diff
@@ -0,0 +1,106 @@
...
+          isSomeday: true,
+        },
+      }),
+    );
+  });
+
+  it("should set correct dates for month category", async () => {
+    const startOfView = dayjs("2024-02-29"); // Leap year February
+    const endOfView = startOfView.add(6, "days"); // Crosses into March but should be ignored
+
```
tyler-dane, line 72: 👌

PR #574, opened 2025-06-30
✨ Feat(web): better UI for migrate event buttons

that-one-arab:
## Description
- use the right arrow and left icons from phosphoricons
- capitalize `CTRL`. This is to keep our casing consistent (We already have SHIFT in another shortcut)
- Add slight margins between icons and title field.

### DIFF
--- a/packages/web/src/views/Forms/EventForm/MigrateBackwardButton.tsx
+++ b/packages/web/src/views/Forms/EventForm/MigrateBackwardButton.tsx
@@ -1,15 +1,8 @@
 import React from "react";
-import styled from "styled-components";
+import { ArrowLeft } from "@phosphor-icons/react";
 import { getMetaKey } from "@web/common/utils/shortcut.util";
 import { Text } from "@web/components/Text";
 import TooltipIconButton from "@web/components/TooltipIconButton/TooltipIconButton";
-
-const StyledMigrateBackwardButton = styled.div`
-  font-size: 25px;
-  &:hover {
-    cursor: pointer;
-  }
-`;
 
 export const MigrateBackwardButton = ({
   onClick,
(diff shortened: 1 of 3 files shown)


that-one-arab:
@tyler-dane I know you only mentioned updating the right arrow but I took the liberty of updating the left arrow too in someday event form and grid event form (move to sidebar button)

I also updated them for someday event rectangle but the new UI was sort of 'eh', LMK your thoughts on it and whether we should keep it

image

tyler-dane:
@that-one-arab I think I was unclear in the original issue about what I was asking for.

**Arrow icon**
Here's a proof-of-concept for what the tooltip should look like.
Screenshot 2025-07-03 at 6 03 30 AM


Notice how the arrow icon is only used for the tooltip. This is because the arrow icon is meant to signify the arrow key (as opposed to the `<` key). 


**Tooltip text**
Notice how the tooltip text does not say "Migrate Backward." We shouldn't use "migrate backward," because that implies that it's going back one week, rather than going to the sidebar. 

If "Migrate to Sidebar" takes up too much space, you could use "To Sidebar" instead

**Event Rectangle Icons**

These should stay `<` and `>`.

These symbols are references to those in the bullet journal migration method, so I'd like to keep them. They're also cleaner than the arrow icons IMO.

https://bulletjournal.com/pages/migration?srsltid=AfmBOooVa0PuwOB-gqv1DYcRFN9Dr1t15_UJBec1qv3yf50yGHBCHkn9

https://bulletjournal.com/blogs/faq/migration?srsltid=AfmBOopAWI4eDXQMmQE0yy6dWp3bQUtO26u0tjOykFtYWOj-FKvIH4Bh

PR #577, opened 2025-07-01
Feat trigger gcal import websocket

### LINKED ISSUE #576: Trigger GCal import via socket events
## Feature Description

We currently trigger the Google calendar full import for a user in a non-deterministic way after a user signs up. 

As a Compass calendar user, 
Once I signup, I am logged in, and currently accessing the Compass calendar UI
- I want my google calendar to be imported into the Compass calendar app.
- I want the import to only happen if I my calendar import has not been previously initiated. 
- I want the import to be triggered only once - even if I am signed in to multiple sessions on different devices. 
- I want the UI to be notified once the calendar import has started.
- I want the UI to be notified once my calendar inport has ended.
- I want this notification to be sent to all my currently active sessions during the calendar import.
So that I can view my google calendar events on the Compass calendar app once they are imported.

## Use Case
Related to #529 - #543 - PR #568 - Deprecating the temporary websocket trigger added [here]((https://github.com/victor-enogwe/compass/blob/ba285c315922fe8eafce419acab92138e451c99c/packages/backend/src/auth/controllers/auth.controller.ts#L185)), in favour of a more deterministic websocket event trigger for the google calendar import function. 

## Additional Context

Hi @tyler-dane I'm adding that feature

> @tyler-dane I'm triggering the sync via a a new backend socket event `IMPORT_GCAL` instead of relying on `UI` state. This event can be called after the user signs up.
> 
> We can:
> 
> * Deterministically send this socket event  `IMPORT_GCAL` after we're sure that the user's socket has connected. This means we need to keep state(`firstLogin` - boolean) on the backend, so this import will not be arbitrarily repeated on successive connections.
> * Non-Deterministically send this socket event  `IMPORT_GCAL` after user sign up, via cron, queue, or staggered event - We'll need to do it within the timeframe we think the user's browser will be open and connected to the socket.
> * Allow the user a manual option to trigger this import once they login(a popup or snack notification can suffice).
> 
> Note that this will be in a separate issue and PR.
> 
> What do you think?

It's related to this comment. The events fetch needs to be triggered. I'll add a temp trigger trigger for now and create another issue to address it fully

_Originally posted by @victor-enogwe in https://github.com/SwitchbackTech/compass/issues/568#issuecomment-3021590634_

victor-enogwe:
## What does this PR do?

This PR
- Adds client websocket event notifications for GCal calendar import process.
- Updates the websocket server to use `supertoken` user `sessions`, so that these events can be triggered based on a user's session

## Description

Closes #576

## Manual Test

**Happy Path**
- Sign up as a new user
- Once you're logged in, confirm that a loading spinning icon appears in the sidebar bottom row
- Hover over the loading spinning icon and confirm that it has a tooltip with the text - "Importing your calendar events in the background"
- Confirm that the spinning loading icon disappears within 30 secs - indicating that the import has completed.


packages/backend/src/common/helpers/common.util.ts
```diff
@@ -1,3 +1,46 @@
...
+  event: Parameters<(typeof emitter)["once"]>["0"],
+  timeoutMs: number = 2000,
+  beforeEvent: () => Promise<unknown> = () => Promise.resolve(),
+  afterEvent: (...args: Payload) => Promise<Result> = (...args) =>
+    Promise.resolve(args as unknown as Result),
+): Promise<Result> {
+  return new Promise((resolve, reject) => {
+    const timeout = setTimeout(() => {
+      clearTimeout(timeout);
+      reject(new Error(`wait for ${String(event)} timed out`));
```
tyler-dane, line 36: For consistency, please create an error object in `common/errors` and then import it and use it here. That'll help us track the expected errors easily

tyler-dane, line 36: For consistency, please create an error object in common/errors and then import it and use it here. That'll help us track the expected errors easily

tyler-dane:
This is looking good! Only other issues is that I see we have a memory leak somewhere in the websocket tests. Can you try to identify the issue and fix it if it's part of this PR? If it's unrelated, we can fix in another one

```
yarn test websocket

//output

Test Suites: 1 passed, 1 total
Tests:       10 passed, 10 total
Snapshots:   0 total
Time:        4.779 s
Ran all test suites matching /websocket/i.
Jest did not exit one second after the test run has completed.

'This usually means that there are asynchronous operations that weren't stopped in your tests. Consider running Jest with `--detectOpenHandles` to troubleshoot this issue.
```

packages/backend/src/common/helpers/common.util.ts
tyler-dane: Since this is the only problem I have with this PR, I'm gonna merge and we can pick this up in the followup PR

PR #586, opened 2025-07-06
feat(web): WIP Synthetically drag someday event to other list when reaching end/beginning of list

### LINKED ISSUE #539: Move someday events up/down between lists while reordering via keyboard
### Feature Description

Allow the user to move a someday event between 'This Week' and 'This Month' lists using only this keyboard flow:

1. `TAB` until they focus on the correct event
2. `Space` to enter into reordering mode
3. `Arrow up/down` until they get the event in the correct location
4. `Space` to stop reordering

This is how it works now. Notice how you can't move the event between the lists using the arrows.

https://github.com/user-attachments/assets/40864efa-bc4e-42d6-b18f-68e42c54f442

### Use Case

This'll save users time. If they're already rearranging an event via the `TAB` -> `SPACE` -> arrow workflow, they would current have to either have to
-  press `SPACE` and then `META + CTRL + ARROW` 
- or click the event and then press `META + CTRL + ARROW`

Both those scenarios are much more cumbersome than continuing to pressing the up/down arrow until the event is in the correct spot.

### Additional Context

Rearranging events via the keyboard is going to be a key value prop for Compass. It'll make the process of adapting your schedule as life changes more flexible and fun, making the app stickier than generic calendar apps.

that-one-arab:
## Description
Meant to close https://github.com/SwitchbackTech/compass/issues/539

This PR is a WIP. Its in a flaky state and has a number of issues that needs fixing

### DIFF
--- a/packages/web/src/views/Calendar/components/Draft/sidebar/hooks/useSidebarActions.ts
+++ b/packages/web/src/views/Calendar/components/Draft/sidebar/hooks/useSidebarActions.ts
@@ -555,6 +555,7 @@ export const useSidebarActions = (
     onPlaceholderClick,
     onSubmit,
     reset,
+    onReorder: reorder,
     resetLocalDraftStateIfNeeded,
     setDraft,
   };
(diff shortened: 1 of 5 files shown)


tyler-dane:
We decided to deprioritize this, as discussed in #539. Closing this PR to ensure that only active PRs show up in the PR list. We can revisit later if/when priorities align

PR #588, opened 2025-07-07
Feature/better ux edit recurring

Amumug:
issue #445


that-one-arab (changes requested):
@Amumug Thanks for the PR!
Unfortunately it has a few problems :(

- You need to uncommit your `.env.example` from the branch history of this PR
- You can still notice that I see the alert popup not the toast from the below screenshot. This happens when I click on a grid event.
- You may need to make the recurring events darker, I dont see any visual difference

image

packages/web/src/common/types/web.event.types.ts
```diff
@@ -25,6 +25,7 @@ export type Schema_SomedayEvent = z.infer<typeof SomedayEventSchema>;
 export interface Schema_GridEvent extends Schema_Event {
   hasFlipped?: boolean;
   isOpen?: boolean;
+  isRecurring: boolean;
```
that-one-arab: We have a computed way of knowing if an event is recurring. Please do not add a new flag to the event schema.

See below image

Image

If you really must add it, you must also refactor these instances as well.

that-one-arab (changes requested):

packages/web/src/common/types/web.event.types.ts
Amumug: I forgot to delete it

PR #617, opened 2025-07-14
✨ feat (WIP): prototype to enhance event creation with optional Google Calendar sync control

### LINKED ISSUE #616: Option to disable syncing newly created events to gcal
### Feature Description

I should have the option to disable syncing newly creating events to google calendar.

This could be a global option (applies to all events) or event level option (checkbox inside the event form) or a combination of both.

### Use Case

Users may not want to sync Compass events back to their google calendar.

### Additional Context

_No response_

that-one-arab:
## Description
Would close https://github.com/SwitchbackTech/compass/issues/616

This is a prototype that demonstrates how this option could be implemented.

Whats missing is the UI part, which is to display an option for the user so they can choose to sync or not sync on their own.


tyler-dane:
Closing this to clean up the PR list so that only active PRs show up there

Issue #584, opened 2025-07-05
Change shortcut for moving a grid event to the sidebar


tyler-dane:
### Feature Description

As a user, I want the shortcut for moving events to the sidebar be consistent with the shortcuts for moving events between the sidebar, so that I don't have to memorize more shortcuts.

In other words, the shortcut to move a timed event (grid or all-day) to the sidebar should be `CTRL + [META] + [LEFT/RIGHT ARROW]`. 

----

We currently have this working for the `[META] + SHIFT + ,` shortcut. Practically, in this issue we just need to change the shortcut to be `[META] + SHIFT + [LEFT ARROW]`.

<img width="836" height="794" alt="Image" src="https://github.com/user-attachments/assets/a3aedd1f-a6df-4b06-bfc3-b5cac21c3f07" />

### Use Case

We're already using the `CTRL + [META]` prefix to move someday events
- `CTRL + [META] + [RIGHT]` when a This Week event is open moves the event to the next week
- `CTRL + [META] + [DOWN]` when a This Week event is open moves the event to the This Month section

Keeping this same pattern for regular events feels intuitive. Users will hopefully internalize the idea that whenever they want to change an event's location, they just need to type `CTRL + [META]` and then play around with the arrows. 

### Additional Context

In the future, we're planning to experiment with allowing the users to reschedule events entirely with the keyboard. For example, they'd go from a sidebar event to the grid and then to the correct time without ever touching the mouse. Having a consistent shortcut experience is important for that workflow to work, so doing this now will pay dividends later.

benjaminguma:
Alright then I'll do that asap

that-one-arab:
@benjaminguma Hey Benjamin! Any updates on this?

benjaminguma:
the changes has been effected, this is the[ link](https://github.com/SwitchbackTech/compass/pull/600) here for the PR

PR #618, opened 2025-07-14
🐛 Fix(web): added hotkeys to move grid events to sidebar ctrl+meta+left

benjaminguma:
@that-one-arab  this should be good

Closes https://github.com/SwitchbackTech/compass/issues/584

### DIFF
--- a/packages/web/src/views/Forms/EventForm/EventForm.tsx
+++ b/packages/web/src/views/Forms/EventForm/EventForm.tsx
@@ -276,7 +276,7 @@ export const EventForm: React.FC<FormProps> = ({
   );
 
   useHotkeys(
-    "meta+ctrl+left",
+    "ctrl+meta+left",
     () => {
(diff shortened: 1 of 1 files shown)


packages/web/src/views/Forms/EventForm/EventForm.tsx
```diff
@@ -286,6 +274,22 @@ export const EventForm: React.FC<FormProps> = ({
     },
     [isFormOpen, onSubmitForm],
   );
+
+  useHotkeys(
+    "meta+ctrl+left",
```
that-one-arab: This does work in its current state, but it should be "ctrl+meta+left" to be consistent with tooltip text and other parts that use the same shortcut combo.

benjaminguma:
changes effected

benjaminguma:
I just tested and I did not experience this please state the steps on how you got this

PR #619, opened 2025-07-15
Feat create dayjs module

### LINKED ISSUE #608: Create Custom `dayjs` module
### Feature Description

_No response_

### Use Case

_No response_

### Additional Context

_No response_

victor-enogwe:
What does this PR do?

This PR adds a customized `dayjs` module for manipulating dates within the Compass app.

## Use Case

closes #608

### DIFF
--- a/packages/core/src/util/date/dayjs.ts
+++ b/packages/core/src/util/date/dayjs.ts
@@ -1,0 +1,73 @@
+import dayjsOriginal, { ConfigType, Dayjs, OptionType } from "dayjs";
+import customParseFormat from "dayjs/plugin/customParseFormat";
+import dayOfYear from "dayjs/plugin/dayOfYear";
+import isBetween from "dayjs/plugin/isBetween";
+import isSameOrAfter from "dayjs/plugin/isSameOrAfter";
+import isSameOrBefore from "dayjs/plugin/isSameOrBefore";
+import timezone from "dayjs/plugin/timezone";
+import utcPlugin from "dayjs/plugin/utc";
+import weekOfYear from "dayjs/plugin/weekOfYear";
+import weekday from "dayjs/plugin/weekday";
+
+export type {
+  ConfigType,
+  ConfigTypeMap,
+  Dayjs,
+  DayjsTimezone,
+  FormatObject,
+  ManipulateType,
+  OptionType,
+  OpUnitType,
+  PluginFunc,
+  QUnitType,
+  UnitType,
+  UnitTypeLong,
+  UnitTypeLongPlural,
+  UnitTypeShort,
+} from "dayjs";
+
+export { Ls } from "dayjs";
+
+dayjsOriginal.extend(utcPlugin);
+dayjsOriginal.extend(timezone);
+dayjsOriginal.extend(weekday);
+dayjsOriginal.extend(dayOfYear);
+dayjsOriginal.extend(weekOfYear);
+dayjsOriginal.extend(isSameOrAfter);
+dayjsOriginal.extend(isSameOrBefore);
+dayjsOriginal.extend(isBetween);
+dayjsOriginal.extend(customParseFormat); // for RFC formatting
+
+export enum DateFormat {
+  RFC5545 = "YYYYMMDD[T]HHmmss[Z]",
+  RFC3339 = "YYYY-MM-DD[T]HH:mm:ss[Z]",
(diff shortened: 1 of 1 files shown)


packages/backend/src/event/services/recur/util/recur.util.ts
```diff
@@ -232,25 +242,16 @@ const _getNextStart = (rule: string, startDate: string, endDate: string) => {
...
-    .second(0);
-
-  const dayOfWeek = date.day();
-
-  let daysUntilNextSunday = (7 - dayOfWeek) % 7;
-  if (daysUntilNextSunday === 0) {
-    daysUntilNextSunday = 7;
-  }
+// could be Sunday or Monday - depending on locale
+const _getFirstDayOfNextWeek = (startDate: string) => {
```
victor-enogwe: Hi @tyler-dane please take a look at this change - this should be the desired logic or not?

packages/backend/src/event/services/recur/util/recur.util.ts
```diff
@@ -232,25 +242,16 @@ const _getNextStart = (rule: string, startDate: string, endDate: string) => {
...
+  const date = dayjs(startDate, YEAR_MONTH_DAY_FORMAT).startOf("d");
 
-  const nextSunday = date.add(daysUntilNextSunday, "day");
-  return nextSunday;
+  return date.add(1, "week").weekday(0);
```
victor-enogwe: specifically check this

packages/backend/.env.example
```diff
@@ -16,6 +16,7 @@ BASEURL=http://localhost:3000/api
 CORS=http://localhost:3000,http://localhost:9080,https://app.yourdomain.com
 LOG_LEVEL=debug         # options: error, warn, info, http, verbose, debug, silly
 NODE_ENV=development    # options: development, production
+TZ="UTC"                # set a timezone for this process - prefer UTC
```
tyler-dane: Is there any drawback from removing this env variable and just defaulting to UTC?

How does allowing a developer to change this value help the user experience?

I'd rather eliminate the complexity of tracking another env variable if possible.

Long-term, we will add TZ support to the UI, so users can adjust it there. For now, we can store events in UTC and display them in the user's time according to their browser

packages/backend/src/event/services/recur/util/recur.util.ts
tyler-dane: I believe so. However, I'm afraid this is scope-creeping. This PR was advertised as creating a simple dayjs wrapper. But here we're attempting to change the business logic without doing any testing. This will not only open us up to bugs, but it'll also make it harder to debug any issues by sneaking it into a seemingly unrelated PR.

Keeping the PRs small and focused on the issue is important, because it prevents us from mixing high-priority work with low-priority work (like this change).

packages/backend/src/event/services/recur/util/recur.util.ts
tyler-dane: See other comment

packages/web/src/common/utils/web.date.util.ts
```diff
@@ -285,7 +281,7 @@ const _getNextWeekInSameMonth = (weekStart: Dayjs) => {
 };
 
 const _getTimeLabel = (date: string) => {
-  const orig = dayjs(date).format(HOURS_AM_FORMAT);
+  const orig = dayjs(date).utc().format(HOURS_AM_FORMAT);
```
tyler-dane: This seems like more scope creep. 

Also, why should we display times labels in the UI in UTC?

victor-enogwe: Nice catch. This should be taken care of within the test by assuming a global time zone.

victor-enogwe:
@tyler-dane  you're right about the scope creep - The initial intention and the naming of this PR introduces the creep.

I'll break this down into separate PRs.

packages/backend/src/event/services/recur/util/recur.util.ts
victor-enogwe: True. This should have unit tests by itself. I'll break this PR down

packages/backend/.env.example
victor-enogwe: I'll leave as is so that the backend can eventually choose whatever time zone it wants via config - For testing and debugging purposes. 

At the moment, the backend default to system timezone - we need it to be consistently chosen and also pick UTC as default - during the `dotenv` parsing and `EnvSchema` validation.

This will be moved to a separate PR also.

Issue #464, opened 2025-05-28
Text is not centered when event is 15 minutes long


tyler-dane:
### Feature Description

The text should always be centered in the rectangle, and it shouldn't appear clipped at the bottom.

<img width="196" alt="Image" src="https://github.com/user-attachments/assets/515b9b5d-2b6f-4c91-a93c-f6860fbdead0" />

### Use Case

This'll give users a consistent UX, rather than leaving them wondering why titles look "off" when the events are short

### Additional Context

If needed, you can shrink the text slightly for these events.

Reference how other calendars handle this scenario, including: Gcal, Vim Cal, Notion Calendar, and Outlook.

dinahmaccodes:
Thanks will make PR soon @tyler-dane

dinahmaccodes:
My apologies for the delay. I'm having some dependency issues as i tried to run the frontend without the backend. I'm going through the documentation and filling in details needed for my .env 
Thank for your patience.

dinahmaccodes:
Hello please can you unassign me? 
I'm having issue doing the setup and it's taking me too long to understand and get to the issue @tyler-dane

that-one-arab:
@dinahmaccodes Sure! If you are having setup issues you can always ask and we will do our best to help. For now I will unassign you.

dinahmaccodes:
Thank you. Is there any telegram or discord group i can use to drop my queries and get help, or should i just drop  my questions here.

tyler-dane:
We don't have a separate chat, so you can just drop them here

dinahmaccodes:
I'm trying to set up the dev environment to work on the text centering UI fix (#464). The setup requires Google Cloud credentials, but GCP - Google Cloud Platform won't let me create an account without adding a payment method (my card recently expired).

I tried using placeholder values in .env which lets the project load, but authentication fails when it tries to connect to Google services.

 Is there a way to run the frontend in development mode without valid Google Cloud credentials? Since this is a pure CSS fix, ideally the frontend should work with mock data.
 I love the idea of this project and would love to contribute to it.

that-one-arab:
@dinahmaccodes Unfortunately there is no way to run the project locally without setting up a Google Cloud 
Platform account :(

sarans-h:
Hii just need clarity the clipped thing is for 15 min and the centered text is for all time range?
@that-one-arab @tyler-dane

tyler-dane:
Unassigned due to inactivity.

[...]

PR #624, opened 2025-07-16
bug/fix-text-alignment-15min-event

Amumug:
issue #464 fixed

### DIFF
--- a/packages/web/src/common/styles/theme.ts
+++ b/packages/web/src/common/styles/theme.ts
@@ -62,7 +62,6 @@ export const theme: DefaultTheme = {
     size: {
       xs: "0.563rem",
       s: "0.688rem",
-      smFluid: "clamp(11px, 1vh, 12px)", // New size
       m: "0.8125rem",
       l: "1rem",
       xl: "1.125rem",
--- a/packages/web/src/common/styles/theme.ts
+++ b/packages/web/src/common/styles/theme.ts
@@ -62,6 +62,7 @@ export const theme: DefaultTheme = {
     size: {
       xs: "0.563rem",
       s: "0.688rem",
+      smFluid: "clamp(11px, 1vw, 12px)", // New size
       m: "0.8125rem",
       l: "1rem",
       xl: "1.125rem",
--- a/packages/web/src/common/styles/theme.ts
+++ b/packages/web/src/common/styles/theme.ts
@@ -62,7 +62,7 @@ export const theme: DefaultTheme = {
     size: {
       xs: "0.563rem",
       s: "0.688rem",
-      smFluid: "clamp(11px, 1vw, 12px)", // New size
+      smFluid: "clamp(8px, 1vw, 12px)", // New size
       m: "0.8125rem",
       l: "1rem",
       xl: "1.125rem",
--- a/packages/web/src/common/styles/theme.ts
+++ b/packages/web/src/common/styles/theme.ts
@@ -62,7 +62,7 @@ export const theme: DefaultTheme = {
     size: {
       xs: "0.563rem",
       s: "0.688rem",
-      smFluid: "clamp(8px, 1vw, 12px)", // New size
+      smFluid: "clamp(11px, 1vh, 12px)", // New size
       m: "0.8125rem",
       l: "1rem",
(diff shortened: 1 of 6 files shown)


packages/web/src/ducks/events/sagas/sync.sagas.ts
```diff
@@ -5,9 +5,8 @@ import { importGCalSlice } from "../slices/sync.slice";
 
 export function* importGCal() {
   try {
-    yield call(SyncApi.importGCal);
-
-    yield put(importGCalSlice.actions.success(undefined));
+    // yield call(SyncApi.importGCal);
+    // yield put(importGCalSlice.actions.success(undefined));
```
that-one-arab: please uncomment these.

packages/web/src/views/Calendar/components/Event/Grid/GridEvent/GridEvent.tsx
```diff
@@ -126,7 +126,7 @@ const _GridEvent = (
         direction={FlexDirections.COLUMN}
         flexWrap={FlexWrap.WRAP}
       >
-        <StyledEventTitle size="m" role="textbox">
+        <StyledEventTitle size="s" role="textbox">
```
that-one-arab: I dont believe making all the text small is the correct way to approach this.

Reference below image to see how google calendar makes small event rectangles text font size smaller.

Image

Basically I am saying that text should be small conditionally and we need to have solid foundation/implementation to conditionally detect if the text is not centered and/or clipping outside the rectangle so we can make it smaller based on that.

that-one-arab (changes requested):

packages/web/src/views/Calendar/components/Event/Grid/GridEvent/GridEvent.tsx
Amumug: for google calendar they used fixed font-size

Amumug: https://github.com/user-attachments/assets/135879b5-e610-4044-b35c-c764045e1673

that-one-arab: @Amumug Youre halfway there. Compare the small event `11px` to the big event which should be larger than `11px`

packages/web/src/common/styles/theme.ts
```diff
@@ -62,6 +62,7 @@ export const theme: DefaultTheme = {
     size: {
       xs: "0.563rem",
       s: "0.688rem",
+      smFluid: "clamp(11px, 1vw, 12px)", // New size
```
that-one-arab: How did you test this? I cannot seem to get the font size of the grid event rectangle text to change based on available space

Amumug: https://github.com/user-attachments/assets/c9d32202-a2e6-4a74-b9f9-975a235d59f9

that-one-arab:
@Amumug  Check this image, you can note that the event on the right has slightly clipped text.

Also, both events are 12px, I expected the right event's text to be smaller in size

image

Amumug:
I don't know how to use different font sizes for events but Google Calendar uses the same font size for all events

that-one-arab:
@Amumug  They do not use the same font size for all events. The font size gets smaller if the event rectangle becomes too small. This is the behaviour we are aiming for.

I think your solution may work if you make sure it applies to the height of the event rectangle instead.

that-one-arab:
@Amumug If you can provide a solution that meets the requirements of the PR then please go ahead! You can also work on something else, its up to you.

Amumug:
Alr cool

that-one-arab:
@Amumug Weird enough I am unable to reproduce this bug anymore on `main`

Can you confirm that you are reproducing it and if yes, show steps?

that-one-arab (changes requested):
unresolved feedback

Amumug:
sorry about that i will review it again

Amumug:
https://github.com/user-attachments/assets/5aaf0ed6-eacb-4e8a-a159-e6468764459a

PR #648, opened 2025-07-22
Refactor randomize test users and db

### LINKED ISSUE #647: Isolate test runs from side effects
### Feature Description

Some tests exhibit flakiness, due to leaking states from other test runs. We want to ensure that all the backend tests run without  side effects and with as little shared state as possible.

### Use Case

Allows us to vrun the test cases in a consistent and predictable manner across different environments.

### Additional Context

_No response_

victor-enogwe:
## What does this PR do?

Fixes the flaky tests.

##Use Case

closes #647


packages/backend/src/__tests__/backend.test.init.ts
```diff
@@ -1,14 +1,10 @@
-import dayjs from "@core/util/date/dayjs";
-import { mockNodeModules } from "@backend/__tests__/helpers/mock.setup";
+const mongoUri = (global as unknown as { __MONGO_URI__: string }).__MONGO_URI__;
 
-process.env["BASEURL"] = "https://foo.yourdomain.app";
+process.env["BASEURL"] = "http;s://foo.yourdomain.app";
```
tyler-dane, line 3: typo @victor-enogwe

packages/core/src/types/event.types.ts
```diff
@@ -30,7 +30,7 @@ export enum Categories_Recurrence {
 
 export type Categories_Recur = "all" | "future";
 
-export type Direction_Migrate = "forward" | "back";
+export type Direction_Migrate = "forward" | "back" | "up" | "down";
```
tyler-dane, line 34: This is good to fix...but we shouldn't sneak in out-of-scope work into massive PRs. It makes it harder to follow

@victor-enogwe

tyler-dane: Disregard, realized this PR never got merged

packages/backend/src/__tests__/backend.test.init.ts
tyler-dane: Disregard, realized this PR never got merged

Issue #621, opened 2025-07-15
Duplicate compass events are being created that reference the same google calendar event id


that-one-arab:
### Where did this happen?

Local code with an up-to-date main branch

### Expected Behavior

When I use my app, I should not see duplicate events that I did not create

### Current Behavior

I am seeing duplicate events being created even though I did not create them, and upon checking them I can see that they reference the same google calendar event ID

Refreshing sometimes adds 1 more duplicate event.

Deleting 1 or 2 duplicate events cause all duplicate events (and original event???) to be deleted.

Duplicate events will continue to happen until the app become unusable.

If sync request is commented out, then this bug stops happening

### Steps to Reproduce
1. Sign into Compass
2. Wait until the user session expires
3. Revisit Compass, using the same browser and user account
5. Notice: The events disappear from the calendar, and Compass backend starts importing your events
6. Refresh the page during the import
7. Wait, while the import and refresh finish
8. Notice: Now there are duplicate events

https://github.com/user-attachments/assets/800978c4-185d-40fe-928c-6a195a4f02e3

<img width="1106" height="831" alt="Image" src="https://github.com/user-attachments/assets/e719c4aa-2a11-4825-a692-0b458e60a17d" />

### Possible Solution
Maybe it is related to gcal background sync or the connection pools

### Context

it essentially brings the app to an unusable state

victor-enogwe:
Hi @that-one-arab I agree.

Keep your current db data intact, and restart the app using a new database.
Go through the sign up steps and play around with the app and see if you can reporduce this.

that-one-arab:
@victor-enogwe @tyler-dane I updated reproduction steps.

victor-enogwe:
Do you see any fail XHR requests?

tyler-dane:
Replicated this today. Added more repro steps and screenshot @victor-enogwe

tyler-dane:
Replicated this again today

tyler-dane:
Moved to In Progress, as this is now affecting prod and should take precedence over other issues @victor-enogwe

PR #666, opened 2025-07-29
🔨 feat(dedupe-sync): add cli script to remove duplicate compass user sync

### LINKED ISSUE #662: Create a Sync Merging Migration Script
### Feature Description

We want to remove duplicate `sync` records per user from the `staging` and `production` deploys, so that Mongo db operations that rely on positional operators are able to run without errors.

### Use Case

This would allow the sync imports to complete successfully for our users in staging and production.

### Additional Context

_No response_

victor-enogwe:
## What does this PR do?

This PR dedupes the users's Compass sync data, preventing the mongo db sync queries from failing

## Use Case

closes #662

### DIFF
--- a/packages/scripts/package.json
+++ b/packages/scripts/package.json
@@ -10,10 +10,12 @@
     "commander": "^10.0.0",
     "dotenv": "^16.0.1",
     "inquirer": "^8.0.0",
+    "lodash.uniqby": "^4.7.0",
     "shelljs": "^0.8.5"
   },
   "devDependencies": {
     "@types/inquirer": "^9.0.1",
+    "@types/lodash.uniqby": "^4.7.9",
     "@types/node": "^20.6.3",
     "@types/shelljs": "^0.8.15",
     "ts-node-dev": "^2.0.0",
--- a/packages/scripts/src/cli.ts
+++ b/packages/scripts/src/cli.ts
@@ -6,6 +6,7 @@ import { CliValidator } from "./cli.validator";
 import { runBuild } from "./commands/build";
 import { startDeleteFlow } from "./commands/delete";
 import { inviteWaitlist } from "./commands/invite";
+import { startSyncDuplicateMergeFlow } from "./commands/merge-sync";
 import { runSeed } from "./commands/seed";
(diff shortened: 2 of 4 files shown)


victor-enogwe:
Follow up @tyler-dane  @that-one-arab  - this is a temporary fix to support local development.

**Approach**

Redesigning the `sync` schema  and queries  to avoid relying on `positional operators` is out of scope at the moment. We rely on this type of query logic in a number of places, and the scope of change is not feasible at the moment. 

This bug appears to be localized, I have not been able to reproduce it in staging or production and there are no duplicate sync records within their respective databases. We should investigate further before starting any extra process to redesign or fix.

**The more permanent solution would involve**:
- The merged PR [here](https://github.com/SwitchbackTech/compass/pull/664) to make the imports atomic.
- Schema changes[Future - Not Planned ATM]
   - adding indexes to the schema
   - Redesigning the schema for safer/more efficient? updates
- Setting up migrations
- Documenting the changes

We will test this further  in staging and production and --
- if it can't be reproduced, this script will be deleted - the commit history will reference it incase anyone wants to refer to it in future.
- if it can still be reproduced, we will plan a fix and deprecate this script after fixing - the commit history will reference it incase anyone wants to refer to it in future.

**Scope**

I had some internal debate about where the script will live. The `private` CI repository did not seem like the proper place for this migration script due to context and it is also not a Js/Ts repository. 

The closest place for the script to live temporarily is within this repository, it has JS/TS support, some cli tools specifically geared towards local development, and this script needs to be where it has issue-context and we can still refer to it even though it will be deleted.

Contributors and regular users who encounter this bug, will have a point of reference to this commit. I'll add this

This commit and issue will also serve as a form of documentation.
We will also mention it within the [troubleshoot](https://docs.compasscalendar.com/docs/guides/troubleshoot) section in the docs

**_The parent issue #621 will be closed and a subsequent PR made to deprecate this cli command as soon as we are done testing, confirming that this bug is localized and can no longer be reproduced within any of our environments._**

tyler-dane:
Fair points, @victor-enogwe. I'm good with your outlined approach.

Issue #667, opened 2025-07-29
Implement onboarding flow


that-one-arab:
### Feature Description

followup to https://github.com/SwitchbackTech/compass/issues/498#issuecomment-3127504267

PR #672, opened 2025-08-03
✨ feat(web): implement onboarding flow

that-one-arab:
## Description

Closes https://github.com/SwitchbackTech/compass/issues/667


packages/web/src/views/Calendar/components/Header/HeaderNote/HeaderNote.tsx
```diff
@@ -99,6 +100,14 @@ export const HeaderNote = forwardRef(
 
     // Load from localStorage on mount
     useEffect(() => {
+      const onboardingNote = OnboardingHeaderNoteLocalStorage.get();
+      if (onboardingNote) {
+        setNote(onboardingNote);
+        localStorage.setItem(STORAGE_KEYS.HEADER_NOTE, onboardingNote);
+        OnboardingHeaderNoteLocalStorage.clear();
+        return;
+      }
```
tyler-dane, line 109: What is the benefit of having a separate storage item for the onboarding note instead of just updating the existing note during onboarding? @that-one-arab

that-one-arab: @tyler-dane My reasoning was that I wanted to add distinctions between the source of the header note and treat a potential case where user may exit the onboarding flow stepper right after they choose a header note which would override the existing one.

tyler-dane: Hmm, I think I see what you mean. 

However, this feels like over-optimization at this point.

If users like this feature enough, we'll store these notes in the DB and separate them by week. 

Having a separate DB document just to track the onboarding header note would be overkill. And tracking the onboarding header note in local storage while the others are stored in the DB is a little confusing.

^ Nothing we need to prioritize now; I'd still like to get more user feedback on this before building it out further But wanted to caution against adding too many layers of complexity into early features like this

Issue #681, opened 2025-08-06
Users invited to waitlist show as false when querying `isOnWaitlist`


that-one-arab:
### Where did this happen?

Staging

### Expected Behavior

When inviting a user to the waitlist by adding them manually, they should show as on waitlist when queried

### Current Behavior

they show up as not on waitlist even though they are.

### Steps to Reproduce

1. Download MongoDB Compass 
1. (Ignore the fact that MongoDB stole our name)
1. Connect to staging environment using the connection string
1. Go to the app (on staging URL -- not prod)
1. Signup for the waitlist (foo@gmail.com)
1. Look at the user's waitlist record
1. Notice: It's status is `waitlisted`
1. Manually change the status to `invited`
1. Make a GET query on staging (not localhost) to check the user's status
- {{stagingUrl}}/api/waitlist?email=foo@gmail.com
1. Note that the field `isOnWaitlist` is false (it should be true)

<img width="1281" height="785" alt="Image" src="https://github.com/user-attachments/assets/e05da265-bde4-4e41-897a-b423d7a8ae9b" />

![Image](https://github.com/user-attachments/assets/9512b898-5f07-433c-bcf8-2dbaaff47647)

### Possible Solution

Check for discrepancies in the env between localhost, staging, and prod. 

### Context

Bug reported by @tyler-dane but authored by me.

tyler-dane:
I'm not able to reproduce this this morning. Will keep an eye on it, though

that-one-arab:
<img width="289" height="539" alt="Image" src="https://github.com/user-attachments/assets/ad8055a9-ba0b-4dbc-9a5f-4c923700323c" />

From what I concluded, it is due to staging mongodb's `dev_calender` and `prod_calendar` databases inconsistencies

I think what added to our confusion, was that we have a `compasscaltest3@gmail.com` in the `prod_calendar.user` but we dont have a waitlist record for them in `prod_calendar.waitlist`, while having a user with the same email in `dev_calendar._dev.waitlist` but not having a record inside `dev_calendar._dev.user`

I'm not sure why we have 2 separate databases for development and production, even though we have dedicated mongodb servers for staging and prod, at this time to me it seems like its making things complicated. If we don't have a good reason for keeping them I suggest we remove this `IS_DEV` logic, check my linked PR.

Regardless, this does not appear to be an issue, we can move on to closing this after you assess and give us the go.

that-one-arab:
[...]

PR #699, opened 2025-08-08
chore(backend): Cleanup dedicated development DB logic

that-one-arab:
This logic is adding extra overhead and making things unnecessarily complicated. Check https://github.com/SwitchbackTech/compass/issues/681#issuecomment-3168717512

This PR is not in a functioning state and should not be merged just yet. We need to verify that the new DB functions properly.

### DIFF
--- a/packages/backend/src/common/constants/env.constants.ts
+++ b/packages/backend/src/common/constants/env.constants.ts
@@ -55,7 +55,7 @@ const processEnv = {
   CHANNEL_EXPIRATION_MIN: process.env["CHANNEL_EXPIRATION_MIN"],
   CLIENT_ID: process.env["CLIENT_ID"],
   CLIENT_SECRET: process.env["CLIENT_SECRET"],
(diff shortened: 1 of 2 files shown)


packages/backend/src/common/constants/env.constants.ts
```diff
@@ -55,7 +55,7 @@ const processEnv = {
   CHANNEL_EXPIRATION_MIN: process.env["CHANNEL_EXPIRATION_MIN"],
   CLIENT_ID: process.env["CLIENT_ID"],
   CLIENT_SECRET: process.env["CLIENT_SECRET"],
-  DB: IS_DEV ? "dev_calendar" : "prod_calendar",
+  DB: "compass_calendar",
```
tyler-dane, line 58: this would be a breaking change unless we included a migration script

tyler-dane:
This strikes me as a nice-to-have if we have time to thoroughly test it. The problem you mentioned in #681 is mitigated now that I cleaned up the excessive datastore in staging

PR #722, opened 2025-08-11
scroll grid to draft location when typing c

### LINKED ISSUE #511: Scroll grid to draft location when typing `C`
### Feature Description

When typing `c` or selecting the `Create Event` action from the command palette, the draft event form should be in focus.

Notice how the form appears, but the user still has to touch the mouse and scroll in order to see it.

https://github.com/user-attachments/assets/0c97396a-4a9a-4982-88d7-c465114593bf

### Use Case

If the user has typed the `c` shortcut or used the command palette, then they don't want to use the mouse. Requiring them to do so would defeat the purpose of having the shortcuts. By scrolling to the form, we make it possible for the user to immediately see their changes and never move from the keyboard.

### Additional Context

We're already scrolling to the Now Line during the first render, so reference that code for an example of how to scroll to an element. See `useScroll.ts`

Amumug:
Closes #511 

https://github.com/user-attachments/assets/eea1ab11-fb38-4b21-94af-d73bce478fb1

### DIFF
--- a/packages/web/src/views/Calendar/hooks/grid/useScroll.ts
+++ b/packages/web/src/views/Calendar/hooks/grid/useScroll.ts
@@ -20,9 +20,29 @@ export const useScroll = (
     }
   }, [timedGridRef]);
 
+  // Scroll when pressing "c"
+  useEffect(() => {
+    const handleKeyDown = (event: KeyboardEvent) => {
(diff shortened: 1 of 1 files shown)


that-one-arab (approved):
lgtm

PR #764, opened 2025-08-19
🧹 Chore: Cleanup `.env.example`

### LINKED ISSUE #763: Comment out optional values in .env.example & remove `SOCKET_USER`
### Feature Description

As a Compass contributor, I want to do minimal configuration before getting the app working, so that I can use it without unnecessary config.

---

We should comment out optional values in `.env.example`, so that we make it extra clear that these values are not needed to run Compass.

We also shouldn't include any superfluous env values. (We're currently including a `SOCKET_USER`, which doesn't have much of a use now that we have unit tests for the socket behavior.

### Use Case

Now that we've added more third-parties to Compass, configuring it can seem intimidating. By leaving all the optional services commented out, we make the initial `.env` setup more approachable.

### Additional Context

_No response_

tyler-dane:
Closes #763

### DIFF
--- a/packages/backend/.env.example
+++ b/packages/backend/.env.example
@@ -65,7 +65,7 @@ SUPERTOKENS_KEY=UNIQUE_KEY_FROM_YOUR_SUPERTOKENS_ACCOUNT
 # PROD_DOMAIN=app.yourdomain.com
 
 ####################################################
-# 6. Email (optional)                              #
+# 6. Email - Kit (optional)                              #
 ####################################################
 # Get these from your Kit.com account
 # If any of these EMAILER_ values are empty, email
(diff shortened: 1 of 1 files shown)


packages/backend/.env.example
```diff
@@ -24,14 +24,12 @@ PORT=3000               # Node.js server
 TOKEN_COMPASS_SYNC=YOUR_UNIQUE_STRING
 TOKEN_GCAL_NOTIFICATION=ANOTHER_UNIQUE_STRING
 
-
 ####################################################
 # 2. Database                                      #
 ####################################################
-MONGOMS_VERSION=8.0.9 # pins the test mongo server binary version
+MONGOMS_VERSION=8.0.9 # pins the test mongo server binary version: https://typegoose.github.io/mongodb-memory-server/docs/api/config-options/#version
```
tyler-dane, line 30: @victor-enogwe Do we need this? I haven't included it in my `.env` and haven't run into any issue. It also seems like the `mongodb-memory-server` defaults to a recent version, which alleviates the need for us to keep this version recommendation updated. 

What is the benefit of pinning a version here?

tyler-dane: If it's not helpful, I'd prefer to remove it from this file to make setup easier for contributors

tyler-dane: Moving forward with this PR, but would still like to know

victor-enogwe: didn't see this on time @tyler-dane 

We should pin the MongoMemoryServer version if we can help it, so the test environments are similar across development machines.

PR #767, opened 2025-08-20
:bug: fix(gcal-import): start watch during resync only when using an https endpoint

### LINKED ISSUE #765: Events are missing after signup if not using Ngrok
### Where did this happen?

### Expected Behavior

When a developer doesn't configure Ngrok, then they should still see their google calendar events after signing up for Compass locally.

Subsequent changes to their Google Calendar events from GCal will not be synced to Compass, which is expected

### Current Behavior

When a developer doesn't configure Ngrok, then they don't see any events on their calendar.

- The `/import-gcal` successfully returns 204 No Content. 
- There is no loading spinner in the left sidebar
- The Gcal events never show up, event after refreshing

### Steps to Reproduce

1. Checkout to `main` and pull
2. Delete any existing test users using the CLI
3. Comment out the NGrok values in `.env`
4. Start web and backend server in separate terminals
5. Go to home: http://localhost:9080
6. Notice: Onboarding flow starts
7. Go through the initial onboarding flow until you finish the Google Oauth
8. Click the Skip Intro button at the "As you know, scallywags..." screen.
9. Calendar renders
10. Expected: Seeing existing Gcal events
11. Actual: No events on calendar

### Possible Solution

Need to make sure that Compass is truly operational without Ngrok configuration. We don't have any runtime errors now, but this is not the expected experience for users. 

### Context

Related to #743 #763

victor-enogwe:
## What does this PR do?

This PR fixes the sync error during local development when the ngrok service which provides an `https` endpoint for the gcal sync is not active.

## Use Case

closes #765

### DIFF
--- a/packages/backend/src/common/services/mongo.service.ts
+++ b/packages/backend/src/common/services/mongo.service.ts
@@ -10,6 +10,7 @@ import {
   ObjectId,
 } from "mongodb";
 import { Logger } from "@core/logger/winston.logger";
+import { Schema_CalendarList as Schema_Calendar } from "@core/types/calendar.types";
 import { Schema_Event } from "@core/types/event.types";
 import { Schema_Sync } from "@core/types/sync.types";
 import { Schema_User } from "@core/types/user.types";
@@ -23,6 +24,7 @@ const logger = Logger("app:mongo.service");
 interface InternalClient {
   db: Db;
   client: MongoClient;
+  calendar: Collection<Schema_Calendar>;
   event: Collection<Omit<Schema_Event, "_id">>;
   sync: Collection<Schema_Sync>;
   user: Collection<Schema_User>;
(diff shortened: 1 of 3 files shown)


packages/backend/src/common/services/mongo.service.ts
```diff
@@ -103,6 +114,7 @@ class MongoService {
     return {
       db,
       client,
+      calendar: db.collection<Schema_Calendar>(Collections.CALENDARLIST),
```
tyler-dane, line 117: I think we should be careful with the naming here. `calendar` and `calendarlist` are two very different things in the Google API.

**calendar**

Represents the calendar itself — its metadata and settings.
	•	Scope: A single calendar (not tied to a specific user by default).
	•	Common uses:
	•	Create a new calendar.
	•	Update a calendar’s name, description, time zone, etc.
	•	Delete a calendar entirely.
	•	Fetch details about any calendar (if you know its ID).

**calendarlist**
Represents the list of calendars that a specific user has access to.
	•	Scope: User-specific — it only shows calendars visible to the authenticated user.
	•	Common uses:
	•	Get all calendars a user can see.
	•	Add an existing calendar to the user’s list.
	•	Update user-specific settings for a calendar (color, selected status, default reminders, etc.).
	•	Remove a calendar from the user’s list without deleting it globally.
	
In general I'm in favor of sticking with the Google API naming convention unless we're adding our own spin on a model, which we're not here yet.

Issue #745, opened 2025-08-16
Sidebar buttons disappear after closing form via `ESC`


tyler-dane:
### Where did this happen?

Local code with an up-to-date main branch, Production app

### Expected Behavior

The sidebar buttons should be visible after closing a sidebar form via `ESC`

### Current Behavior

The sidebar buttons disappear after closing a sidebar form via `ESC`

https://github.com/user-attachments/assets/70e6cd04-ff04-42ba-99d0-fa3931c52867

### Steps to Reproduce

1. Sign in to compass
2. Open a someday form by typing `W` or `M` (click the buttons, must use shortcut)
3. Close form by typing `ESC`
4. Expected: the "+" buttons in the sidebar are visible
5. Actual: the buttons are gone now 
6. Type `W` or `M` and notice that the forms appear as expected, even if buttons are invisible.
7. Refresh page
8. Buttons appear again

### Context

This is a confusing experience for users and discourages them from using the keyboard to manage their calendar. 

Being keyboard-friendly is a major differentiator for us. Although there are workarounds for this, this is still a higher priority issue because it affects one of the major use-cases for the app (creating events)

Amumug:
@tyler-dane my main branch updated but i don't have this bug 

https://github.com/user-attachments/assets/fd9ea4f4-0db3-4ced-946c-91253bc98467

PR #888, opened 2025-09-08
🐛 fix(web): opening & closing someday event draft form using shortcuts leaves hanging draft state

that-one-arab:
## Description
Closes https://github.com/SwitchbackTech/compass/issues/745

- Centralize create someday draft logic between the shortcut event and the placeholder button event
- Simplify create someday draft logic
- Handle few TS errors along the way

We still need to centralize logic for creating someday draft in the command palette component too, but needed to get to a stopping point.

### DIFF
--- a/packages/web/src/common/utils/draft/someday.draft.util.ts
+++ b/packages/web/src/common/utils/draft/someday.draft.util.ts
@@ -6,6 +6,10 @@ import { draftSlice } from "@web/ducks/events/slices/draft.slice";
 import { Activity_DraftEvent } from "@web/ducks/events/slices/draft.slice.types";
 import { assembleDefaultEvent } from "../event.util";
 
+/** @deprecated
+ * `createSomedayDraft` should not be called outside `useSidebarActions.createSomedayDraft`
+ * Consider removing
+ */
 export const createSomedayDraft = async (
   category: Categories_Event.SOMEDAY_WEEK | Categories_Event.SOMEDAY_MONTH,
   startOfView: Dayjs,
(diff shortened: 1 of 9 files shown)


packages/web/src/common/utils/draft/someday.draft.util.ts
```diff
@@ -6,6 +6,10 @@ import { draftSlice } from "@web/ducks/events/slices/draft.slice";
 import { Activity_DraftEvent } from "@web/ducks/events/slices/draft.slice.types";
 import { assembleDefaultEvent } from "../event.util";
 
+/** @deprecated
+ * `createSomedayDraft` should not be called outside `useSidebarActions.createSomedayDraft`
+ * Consider removing
+ */
```
tyler-dane, line 12: What is your plan with this?

Adding a vague comment like this and not tracking it in an issue is not production-quality. We don't use comments as task management

packages/web/src/views/Calendar/components/Draft/sidebar/hooks/useSidebarState.ts
```diff
@@ -22,7 +22,7 @@ export const useSidebarState = (measurements: Measurements_Grid) => {
   const [isDraftingExisting, setIsDraftingExisting] = useState(false);
   const [isSomedayFormOpen, setIsSomedayFormOpen] = useState(false);
 
-  const isDragging = isDrafting && draft !== null;
+  const isDragging = isDrafting && draft !== null; // TODO: Probably not a good way to determine if we are dragging, consider refactoring or removing this comment.
```
tyler-dane, line 25: What is your plan with this?

Adding a vague comment like this and not tracking it in an issue is not production-quality. We don't use comments as task management

PR #973, opened 2025-09-24
:bug: fix(calendar-event-transiton): calendar event to someday fix

victor-enogwe:
## What does this PR do?

this PR fixes the calendar events to someday events transition bug.

## Use Case

closes #924


packages/core/src/types/event.types.ts
```diff
@@ -150,13 +150,17 @@ export enum CompassEventStatus {
   CANCELLED = "CANCELLED",
 }
 
+export const idSchema = z.string().refine(ObjectId.isValid, {
+  message: "Invalid id",
+});
```
tyler-dane, line 155: 🔥

PR #1027, opened 2025-09-29
:sparkles: feat(delete-recurrence): allow deletion of  provider orphaned compass event

### LINKED ISSUE #1026: Allow deletion of provider-orphaned events from Compass when the corresponding Google Calendar event no longer exists
### Priority

High (would significantly improve my experience)

### Feature Description

Currently, when a user attempts to delete an event in Compass that has already been deleted from the Google Calendar provider (but the corresponding sync event was missed), Compass attempts to delete the event from Google Calendar again. Since the event no longer exists in Google Calendar, this results in an error and prevents the deletion from proceeding in Compass.

This feature will allow the delete operation to proceed on Compass if the corresponding event no longer exists in the Google Calendar, removing the orphaned event from the Compass calendar.

### Use Case

Users expect that deleting an event from their Compass calendar should always succeed, even if the event has already been deleted from the external provider. 

This change will improve user experience by preventing confusing errors and ensuring calendar consistency, especially in cases where sync events are missed or delayed.

### Additional Context

- Sync events from providers may be missed due to network issues or API limitations, resulting in orphaned events.
- The Compass backend currently retries provider deletes and fails if the event is not found.
- The solution should detect provider "not found" errors and treat them as successful deletes for Compass.

victor-enogwe:
## What does this PR do?

This PR enables the delete operation to proceed on Compass if the corresponding event no longer exists in the Google Calendar, removing the orphaned event from the Compass calendar.

## Use Case

closes  #1026

### DIFF
--- a/packages/backend/src/event/services/event.service.ts
+++ b/packages/backend/src/event/services/event.service.ts
@@ -190,33 +190,39 @@ class EventService {
       .find({ user: userId, _id: { $in: baseEventIds } })
       .toArray();
 
-    return events.map((event) => {
-      if (isExistingInstance(event)) {
-        const baseEvent = baseEvents.find(
-          ({ _id }) => _id.toString() === event.recurrence?.eventId,
-        );
-
-        if (!baseEvent) {
-          throw new BaseError(
-            "Base event not found for instance",
-            `Tried with user: ${userId} and _id: ${event._id.toString()}`,
-            Status.NOT_FOUND,
-            true,
+    return events
+      .map((event) => {
+        if (isExistingInstance(event)) {
+          const baseEvent = baseEvents.find(
+            ({ _id }) => _id.toString() === event.recurrence?.eventId,
           );
+
+          if (!baseEvent) {
+            console.error(
+              new BaseError(
+                "Skipping instance. Base event not found for instance",
+                `Tried with user: ${userId} and _id: ${event._id.toString()}`,
+                Status.NOT_FOUND,
+                true,
+              ),
+            );
+
+            return undefined;
+          }
+
+          return {
+            ...event,
+            _id: event._id.toString(),
+            recurrence: {
+              eventId: baseEvent._id.toString(),
+              rule: baseEvent.recurrence?.rule,
+            },
+          };
         }
 
-        return {
-          ...event,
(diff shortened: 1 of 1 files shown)


packages/backend/src/event/services/event.service.ts
```diff
@@ -489,13 +507,24 @@ export const _deleteGcal = async (
...
+    }
+
+    const response = await gcalService.deleteEvent(gcal, gEventId);
+
+    return response.status < 300;
+  } catch (e) {
+    const error = e as GaxiosError<gSchema$Event>;
 
-  const response = await gcalService.deleteEvent(gcal, gEventId);
+    if (error.code?.toString() === "410") return true;
```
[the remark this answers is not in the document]
victor-enogwe: error.code is typed as a number in the "gaxios" library, but is returned as a string. We need to make this conversion to ensure a string is always returned.

packages/backend/src/event/services/event.service.ts
```diff
@@ -235,8 +250,17 @@ class EventService {
         _id: new ObjectId(event.recurrence?.eventId),
       });
 
+      if (!baseEvent) {
+        throw new BaseError(
+          "Base event not found for instance",
```
tyler-dane, line 255: I think we should use `error()` for consistency, rather than having some places `throw Error` and others `throw error()`.

packages/backend/src/event/services/event.service.ts
tyler-dane, line 1: Would've like to see tests to this business logic change

victor-enogwe: Same here, it's a bit jejune to mock this Gcal behaviour. We'll figure something out.

packages/backend/src/event/services/event.service.ts
victor-enogwe: It's following existing code patterns in this file. We should do a refactor effort solely for standardizing errors at some point.

tyler-dane: We can make that refactor effort easier by sticking to one standard along the way, like using `error()`. The fact that there are currently multiple patterns in `event.service` shouldn't deter us from cleaning up along the way

PR #1062, opened 2025-10-05
fix(web): handle someday week label edge case

### LINKED ISSUE #1035: Incorrect Date Label During Month Changes
### Where did this happen?

Hosted (app.compasscalendar.com)

### Expected Behavior

Label should show `This Week` 

### Current Behavior

Label shows day labels

Image

### Steps to Reproduce

**Option 1: Wait**
Wait until the current day is on a month change

**Option 2: Recreate**
Change the current date to simulate the problem. 

1. Update this code in `useToday.ts`:

```
//useToday.ts

export const useToday = () => {
  const getToday = (todayIndex: number) => {
    return dayjs("2025-09-03"); // 👈 hard-coded to this date to replicate the bug
    const today = dayjs();
    if (today.get("day") === todayIndex) {
      return today;
    }
    return today;
  };
```

2. Restart the web and backend servers
3. Look at the Week someday section

### Possible Solution

Make the day labels smarter:
- Add a function called `getSomedayWeekLabel()`, which adapts the code in `WeekSection.tsx:isCurrentWeek()` so that it supports this edge-case
- Export the business logic from ^ and write a unit test to validate its functionality outside of the component
- Use `getSomedayWeekLabel()` in `WeekSection`

### Context

**Troubleshooting**: Artificially changing the date like this might cause the session management code to get out-of-whack, which will probably result in having to log in more frequently. If this gets annoying, consider commenting out the `Supertokens` session provider in `App.tsx` and `express.server.ts` while developing. Just remember to uncomment that code when submitting your PR for review.

potoky:
Fixes #1035 

**🐞 Problem**
The existing isCurrentWeek function:
- Did not show title "This Week" when the day was within the range of this week.
- Inflexible — couldn’t handle labels like "12.29 - 1.4" (cross-year) or "10.4 - 6" (short label format).

**💡 Impemented**
- [x] Implemented correct calculation of this week range and comparsion with actual day.
- [x] Improved year rollover detection for cross-year and cross-month weeks.
- [x] Implemented tests - see below.

**🧪 Testing**
Added new test cases covering:
- [x] Current week
- [x] Week immediately before and after the current one
- [x] Weeks that cross New Year and month boundaries
- [x] Partial labels like "10.4 - 6"

Verified that:
- The function correctly returns "This Week" when the current date falls within the label.
- It returns the week label otherwise.

this_week


packages/web/src/views/Calendar/components/Sidebar/SomedayTab/WeekSection/isCurrentWeek.test.tsx
```diff
@@ -0,0 +1,225 @@
...
+  const startStr = `${start.getMonth() + 1}.${start.getDate()}`;
+  const endStr = `${end.getMonth() + 1}.${end.getDate()}`;
+  return `${startStr} - ${endStr}`;
+}
+
+export function getViewStart(offset: number = -3): Date {
+  return dayjs().add(offset, "day").startOf("day").toDate();
+}
+
+describe("WeekSection Component", () => {
```
tyler-dane: Since the WeekSection component doesn't actually do anything with the label value aside from displaying it, I don't think the tests in this `WeekSection Component` `describe` block provide any extra value. In fact, the helper function and mocks just make it more difficult to maintain. 

Let's remove those and focus on fixing the `isCurrentWeek` / `getWeekLabel` functions and their tests instead

packages/web/src/views/Calendar/components/Sidebar/SomedayTab/WeekSection/WeekSection.tsx
```diff
@@ -19,38 +20,49 @@ interface Props {
...
-
   return (
     <SidebarSection>
       <SidebarHeader
         alignItems={AlignItems.CENTER}
         justifyContent={JustifyContent.SPACE_BETWEEN}
       >
         <Text role="heading" size="xl">
-          {isCurrentWeek(weekLabel) ? "This Week" : weekLabel}
+          {isCurrentWeek(weekLabel, viewStart) ? "This Week" : weekLabel}
```
tyler-dane: Simply checking whether it's the current week isn't sufficient to fix the edge case listed in #1035, because that doesn't account for when it's a month change.

tyler-dane (changes requested):

packages/web/src/views/Calendar/components/Sidebar/SomedayTab/WeekSection/WeekSection.tsx
```diff
@@ -17,40 +18,67 @@ interface Props {
...
+  const startDate = new Date(startYear, startMonth - 1, startDay);
+  const endDate = new Date(endYear, endMonth - 1, endDay);
+
+  today.toDate().setHours(0, 0, 0, 0);
+  startDate.setHours(0, 0, 0, 0);
+  endDate.setHours(0, 0, 0, 0);
+
+  console.log(startDate + " " + today + " " + endDate);
+
+  return today.toDate() >= startDate && today.toDate() <= endDate;
```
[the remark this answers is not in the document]
tyler-dane: Let's get away from handling so much `Date` manipulation by hand, which can be prone to these types of bugs. Instead, please use `dayjs`

packages/web/src/views/Calendar/components/Sidebar/Sidebar.tsx
tyler-dane, line 1: Instead of prop drilling, please just use the `useToday` hook where it's needed

tyler-dane (approved):
Made a few quick tweaks, hope you don't mind. Congrats on the first PR! 🎉

Issue #1060, opened 2025-10-05
Improve accessibility and styles for event form when editing with keyboard


tyler-dane:
## Motivation

Editing event via keyboard should be fast and follow standard accessibility best-practices, so that users can quickly make calendar changes while staying in the flow

## Requirements

General
- All elements have a style that changes when focused
- All form elements have a style that changes when hovered

## Context
This is a follow-up to #934

WanyGM:
Hi! I'm a student in the CodePath AI301 course, and I'd like to work on this issue. I'm comfortable with React and TypeScript, and I can start by reproducing the problem locally this week. Any pointers on where to start in the codebase would be appreciated!

Issue #1076, opened 2025-10-07
Improve Accessibility: Date Picker Keyboard and Screen Reader Interactions


tyler-dane:
### Priority

None

### Feature Description

Ensure the date picker trigger and calendar are fully accessible by keyboard, with correct roles (dialog, grid, gridcell), focus and date selection controls, and ARIA attributes. Support keyboard navigation across days/months/years and proper announcement of the selected date. Follows ARIA Dialog, Grid, and GOV.UK accessibility guidance. Motivation: Enables all users to efficiently select dates without leaving the keyboard, critical for inclusivity.

## Spec
  - Trigger button (aria-haspopup="dialog") gains focus via Tab; Enter/Space open the calendar and move focus to the selected (or today’s) date.
  - Calendar uses role="grid" with month/year label; days are role="gridcell" with aria-selected and today flagged via aria-current="date".
  - Arrow keys move by day; Home/End jump to week bounds; PageUp/PageDown change month; Shift+PageUp/PageDown change year while respecting limits.
  - Enter/Space select the focused date, close the popover, and return focus to the trigger with the new date announced.
  - Escape or outside click closes without changes; Tab/Shift+Tab from the popover edges close it and continue the document’s tab order.

### Use Case

Allows users—especially those with disabilities—to select dates confidently and efficiently using keyboard or screen reader, reducing friction and supporting accessibility conformance.

### Additional Context

References:
- WAI-ARIA APG – Dialog (Modal) Pattern https://www.w3.org/WAI/ARIA/apg/patterns/dialog-modal/
- WAI-ARIA APG – Grid Practice https://www.w3.org/WAI/ARIA/apg/patterns/grid/
- GOV.UK Design System – Date Inputs Accessibility https://design-system.service.gov.uk/components/date-input/accessibility/

### Contribution Interest

- [ ] I would like to help implement this feature
- [ ] I can help with design/mockups
- [ ] I can help with testing
- [ ] I can help with documentation
- [ ] I'm just suggesting, not planning to contribute

PR #1088, opened 2025-10-08
feat(web): Improve accessibility and styles for event form when editing with keyboard

PatrikH-01:
### Changes
Added :hover and :focus effects to elements in EventForm that didn't have them. I also improved styling of IntervallInput and TimePicker components a little bit.

Before, after comparison:
image

Here is a video of how the form looks after styling changes:
https://github.com/user-attachments/assets/bb5adb19-5ed3-4408-bbd8-005ec9f88114

### Related Issue
Related to #1060

### DIFF
--- a/packages/web/src/components/DatePicker/DatePicker.tsx
+++ b/packages/web/src/components/DatePicker/DatePicker.tsx
@@ -2,7 +2,7 @@ import classNames from "classnames";
 import React, { useEffect, useRef } from "react";
 import ReactDatePicker, { ReactDatePickerProps } from "react-datepicker";
 import "react-datepicker/dist/react-datepicker.css";
-import { darken, hslToSpaceFormat, isDark } from "@core/util/color.utils";
+import { darken, isDark } from "@core/util/color.utils";
 import dayjs from "@core/util/date/dayjs";
 import { theme } from "@web/common/styles/theme";
 import {
@@ -84,7 +84,7 @@ export const DatePicker: React.FC<Props> = ({
       customInput={
         <Focusable
           Component={StyledInput}
-          underlineColor={hslToSpaceFormat(darken(bgColor, -15))}
+          underlineColor={darken(bgColor, -15)}
           bgColor={inputColor}
           withUnderline
         />
--- a/packages/core/src/util/color.utils.ts
+++ b/packages/core/src/util/color.utils.ts
@@ -11,11 +11,3 @@ export const darken = (color: string, amount?: number) => {
 export const isDark = (color: string) => {
   return tinycolor(color).isDark();
 };
-
-export const hslToSpaceFormat = (hsl: string) => {
-  return hsl
-    .replace(/,/g, "")
-    .replace(/%/g, "")
-    .replace(/\s+/g, " ")
-    .replace(/hsl\s*\((.*)\)/, "hsl($1)");
-};


packages/web/src/components/ContextMenu/styled.ts
```diff
@@ -46,7 +46,16 @@ export const PriorityCircle = styled.div<{ color: string; selected: boolean }>`
   background-color: ${({ selected, color }) =>
     selected ? color : "transparent"};
   cursor: pointer;
-  transition: all 0.2s ease-in-out;
+  transition: ${({ theme }) => theme.transition.default};
+
+  &:hover {
+    background-color: rgba(0, 0, 0, 0.1);
```
tyler-dane: Please don't add hardcoded colors. Instead, use the theme.

Including them seems innocent enough here, but avoiding this will make our lives much easier once we get to #1094

packages/web/src/views/Forms/EventForm/DateControlsSection/DateTimeSection/DateTimeSection.tsx
```diff
@@ -69,6 +71,7 @@ export const DateTimeSection: FC<Props> = ({
           setSelectedStartDate={setSelectedStartDate}
           setIsEndDatePickerOpen={setIsEndDatePickerOpen}
           setIsStartDatePickerOpen={setIsStartDatePickerOpen}
+          underlineColor={underlineColor}
         />
```
tyler-dane, line 72: Adding another prop to this already massive component and drilling it a lot with `underlineColor` is making me wary. 

Also...I'm not sure what it's actually doing. I set it to "red" and don't see any changes in the UI. 

Maybe I'm missing something, but in general I feel like we can make things simpler

packages/web/src/views/Forms/EventForm/styled.ts
```diff
@@ -31,9 +32,32 @@ export const StyledDescription = styled(Textarea)`
...
+    background: ${({ theme }) => theme.color.border.primaryDark};
+    border-radius: 999px;
+  }
+
+  @-moz-document url-prefix() {
+    & {
+      scrollbar-width: thin;
+      scrollbar-color: ${({ theme }) => theme.color.border.primaryDark}
+        transparent;
+    }
```
tyler-dane, line 63: Another source of wariness here. 

Since the scrollbar styles are working fine in `main`, the fact that we need to do so much on this branch to preserve that same styling makes me think this is more of a bandaid that is covering up some other styling issue, rather than code that we should keep around

PatrikH-01: Thanks for pointing this out. That’s interesting, on my environment (Windows + Chrome), the scrollbar doesn’t look correct even on main branch, which is why I added the extra CSS. Could you let me know what OS/browser you’re using where it already looks correct?

If it’s just an environment difference, maybe we can add a cross-browser compatible style (scrollbar-color for Firefox, ::-webkit-scrollbar for Chromium browsers) as i already did ant that should ensure consistent looks everywhere.

tyler-dane: Oh interesting. I'm on Mac + Chrome. 

If `main` isn't working on windows then I agree, let's go with your working version.

packages/web/src/components/DatePicker/DatePicker.tsx
```diff
@@ -80,7 +81,14 @@ export const DatePicker: React.FC<Props> = ({
           view={view}
         />
       )}
-      customInput={<StyledInput bgColor={inputColor} />}
+      customInput={
+        <Focusable
+          Component={StyledInput}
+          underlineColor={darken(bgColor, -15)}
```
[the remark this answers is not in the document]
tyler-dane: Will be addressed in #1076

PR #1469, opened 2026-02-20
feat(auth): implement session user synchronization and related tests

### LINKED ISSUE #1467: Tasks still store `"UNAUTHENTICATED_USER"` after connecting gcal
### Where did this happen?

_No response_

### Expected Behavior

After a user connects their gcal, their tasks in IDB should be updated so that the user is the user's id from MongoDB

### Current Behavior

_No response_

### Steps to Reproduce

_No response_

### Possible Solution

_No response_

### Context

_No response_

tyler-dane:
- Added `syncLocalTaskUsersToAuthenticatedUser` utility to synchronize task users with the authenticated user upon login.
- Updated `SessionProvider` to trigger synchronization when the authentication state changes.
- Introduced tests for `SessionProvider` to verify synchronization behavior based on authentication status.
- Enhanced `LocalTaskRepository` to normalize unauthenticated users when saving tasks.
- Added comprehensive tests for the synchronization utility to ensure correct behavior across various scenarios.
Closes #1467

### DIFF
--- a/packages/web/src/auth/session/SessionProvider.test.tsx
+++ b/packages/web/src/auth/session/SessionProvider.test.tsx
@@ -1,0 +1,63 @@
+import { useContext } from "react";
+import { fireEvent, render, screen, waitFor } from "@testing-library/react";
+import { syncLocalTaskUsersToAuthenticatedUser } from "@web/common/utils/sync/local-task-user-sync.util";
+import { SessionContext, SessionProvider } from "./SessionProvider";
+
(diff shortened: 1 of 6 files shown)


tyler-dane:
Closing as unnecessary for now. We can follow the same pattern that we're using for events for now:

When a user connects their GCal, we convert IDB events to cloud events. In other words, they're deleted from IDB. This is a simple way to move forward.
If/when we want to support full offline mode, we'll want to revisit this, so that users can switch between IDB and cloud events seamlessly.

PR #1756, opened 2026-05-14
https://github.com/HirenGajjar/compass/compare/fix/now-view-esc-key

### LINKED ISSUE #1751: `ESC` does not work for /now
### Where did this happen?

Hosted (app.compasscalendar.com)

### Expected Behavior

When a user types `ESC` while focused on a task in `/now` view, it should redirect user to the `/day` view

### Current Behavior

Nothing happens when typing ESC

### Steps to Reproduce

1. Go to /day
2. Create some tasks for the current day
3. Go to /now
4. Type ESC

### Possible Solution

Add an ESC handler, following the existing pattern 

### Context

Since the now view is immersive, it feels like a full-screen dialog, which makes users instinctively think they can escape out of it.

HirenGajjar:
Fixes #1751

Added ESC key handler to the /now view that navigates the user back to /day.

- Added `onEscape` prop to `useNowShortcuts`
- Added `handleEscape` callback in `NowViewProvider` using `navigate(ROOT_ROUTES.DAY)`
- Added test for the new shortcut


tyler-dane:
Hey @HirenGajjar, thanks for taking the time to submit a PR. Unfortunately, I'm going to close it due to the dependency bloat it'd introduce.

PR #1760, opened 2026-05-15
fix(self-host): allow password-only config refresh

Uarmagan:
## Summary

This is a tiny follow-up for PR #1755. The YAML config work correctly makes Google optional, and the config docs say `google.notificationToken` is only required for HTTPS Google webhooks. The installer refresh validator was stricter than that: any existing `compass.yaml` without `google.notificationToken` failed validation before refresh, even for password-only installs.

## Fix

- Keep required secret checks for Mongo, SuperTokens, and the Compass sync token.
- Change the Google notification-token check to validate the token only when it is present.
- Add a shell-backed regression test showing a password-only config passes, while the old placeholder token is still rejected.

## Verification

- `bun test self-host/docker-compose.test.ts`
- `sh -n self-host/install.sh`
- `sh -n self-host/compass`
- `bun run lint`
- `bun run type-check`

### DIFF
--- a/self-host/docker-compose.test.ts
+++ b/self-host/docker-compose.test.ts
@@ -1,11 +1,66 @@
 import { describe, expect, it } from "bun:test";
-import { readFileSync } from "node:fs";
+import { spawnSync } from "node:child_process";
+import { mkdtempSync, readFileSync, rmSync, writeFileSync } from "node:fs";
+import { tmpdir } from "node:os";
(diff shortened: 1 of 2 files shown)


tyler-dane:
Closing in favor of the simpler solution from https://github.com/SwitchbackTech/compass/pull/1755/changes/2aa153796d18dad4e9467f0ae02b8b87952005aa

PR #1787, opened 2026-05-19
fix(web): prevent no-op drag form

### LINKED ISSUE #1785: Cursor does not transition to `move` after mouse is held down
### Where did this happen?

Hosted (compasscalendar.com)

### Expected Behavior

If the user presses down on a calendar event and does not move the mouse while holding it down, the cursor should transition to `move` and the event should transition to become a placeholder. In other words, it should work like in Google Calendar.

### Current Behavior

Cursor stays a `default`

### Steps to Reproduce

1. Create an event
2. Mouse down on event
3. Hold mouse down long enough for cursor to change
4. Cursor does not change

### Possible Solution

_No response_

### Context

Related to #229

The current UX attempts to handle this situation as a "click" from the user's perspective. However, it's not a click — it's a drag. The user moused down, waited long enough for the cursor to change from `default` to `move` and then moused up without changing the times. This should be handled like any other drag event, where the form does not open afterwards. Behind the scenes, we might consider diffing and returning early if there was no change to the start dates to avoid an API call.

zhouzhou626:
## Summary
- treat no-op saved drag/resize commits as no action when the event form was closed before motion started
- preserve the existing behavior of restoring the form if it had been open before the interaction
- add focused coverage for the no-op form restoration decision

Fixes #1785

## Tests
- `bun install --frozen-lockfile`
- `bun test --cwd packages/web src/views/Week/interaction/WeekInteractionCoordinator.test.ts src/views/Week/interaction/adapter/WeekInteractionAdapter.timedDrag.test.ts`
- `tsc --noEmit`
- `tsc -p packages/web/tsconfig.app.json --noEmit`
- `tsc -p packages/web/tsconfig.test.json --noEmit`
- `bun packages/scripts/src/testing/run.ts web`

Note: `bun run lint` still reports existing issues in `self-host/docker-compose.test.ts` and `packages/web/src/views/Week/interaction/WeekPointerCaptureBoundary.tsx`; the touched files pass `biome check`.

### DIFF
--- a/packages/web/src/views/Week/interaction/WeekInteractionCoordinator.test.ts
+++ b/packages/web/src/views/Week/interaction/WeekInteractionCoordinator.test.ts
@@ -1,0 +1,31 @@
+import { shouldReopenFormAfterNoopSavedMutation } from "./WeekInteractionCoordinator";
+import { describe, expect, it } from "bun:test";
+
+describe("shouldReopenFormAfterNoopSavedMutation", () => {
+  it("does not open the form for a no-op drag that started from the closed form state", () => {
+    expect(
+      shouldReopenFormAfterNoopSavedMutation({
+        hadFormOpenBeforeInteraction: false,
+        hasMoved: false,
+      }),
+    ).toBe(false);
+  });
+
+  it("reopens the form after a no-op interaction if it was open before motion started", () => {
+    expect(
+      shouldReopenFormAfterNoopSavedMutation({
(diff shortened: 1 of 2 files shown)


tyler-dane:
Bug has been updated to reflect the current behavior. The original bug was fixed in a recent PR

tyler-dane:
Thank you for the PR, however, I'm closing this due to the bug's behavior shifting

PR #1799, opened 2026-05-20
fix: use info color for anonymous signup icon

### LINKED ISSUE #1775: Inconsistent icon colors for anonymous users
Image

The top icon should use the blue accent color for consistency

jaryalarshita:
## Summary

* Updated the anonymous user header icon to use the info/accent color instead of the warning color
* Improved visual consistency between anonymous user indicators across the UI
* Updated related test description wording to reflect the new icon semantics

image

## Issue

Closes #1775


packages/web/src/auth/google/hooks/useConnectGoogle/useConnectGoogle.types.ts
```diff
@@ -4,7 +4,7 @@ export type GoogleUiState = "checking" | "repairing" | GoogleConnectionState;
 
 export type CommandActionIcon = "CloudArrowUpIcon";
 
-export type IconColor = "muted" | "warning" | "error";
+export type IconColor = "muted" | "warning" | "error" | "info";
```
tyler-dane: instead of adding complexity by including another type, let's use this as an opportunity to clean up. This `IconColor` is only used in one place. 

Please remove the `IconColor` and `ICON_COLOR_MAP` to simplify how the color is passed to the component. No need to add so much logic for just one color

tyler-dane (changes requested):

packages/web/src/auth/google/hooks/useConnectGoogle/useConnectGoogle.types.ts
jaryalarshita: I’ve made the requested changes and cleaned up the implementation. Could you please take another look when you get a chance?

tyler-dane: looks good, thx

tyler-dane (approved):

PR #1807, opened 2026-05-21
fix(web): clean up now view shortcuts — labels, key sequences, and platform tooltip

### LINKED ISSUE #1791: Shortcut labels are confusing
### Remove Logout
- Remove `Logout` shortcut from the day view and week view. It's already in the command palette, and so we can keep this list shorter and more focused.

### Toggle -> Show
- Rename 'Toggle shortcuts' to 'Show shortcuts.' "Toggle" is ambiguous, while "show" is very clear. Since the user will only see that message when the shortcut sidebar is open, we can use "show" confidently

### Toggle -> Close
- Rename 'Toggle sidebar' to 'Close sidebar'

### Edits on Now
- Change `e` to `e` `d` for editing the description
- Change `r` to `e` `r` for editing the reminder
- Move the reminder from the Global list to the Now list

### Hover over save on now
- When hovering over the save icon on the now view, it shows `Mod+Enter`. It should show the cmd icon on mac and control on windows

HirenGajjar:
## Summary
- Replaces the hacky `flushSync` / `useLayoutEffect` Now description workaround with the targeted Escape hotkey guard.
- Keeps the Now global Escape shortcut active outside inputs while ignoring Escape from textarea/input-like editing contexts.
- Adds a regression test for Escape inside a focused textarea and does small cleanup around the description shortcut label/handlers.

## Root Cause
TanStack hotkeys does not ignore input elements for Escape by default, so the Now view global Escape handler could run while the task description textarea was focused. That navigated to Day instead of letting `TaskDescription` cancel/blur the edit.

## Validation
- `bun lint`
- `bun test ./src/views/Now/shortcuts/useNowShortcuts.test.tsx ./src/views/Now/components/TaskDescription/TaskDescription.test.tsx`
- `bunx typescript@6.0.3 -p packages/web/tsconfig.test.json --noEmit`
- `bun test:web`
- `bunx react-doctor@latest . --verbose --diff` → 98 / 100


packages/web/src/views/Now/components/TaskDescription/TaskDescription.tsx
```diff
@@ -242,7 +242,14 @@ export const TaskDescription: React.FC<TaskDescriptionProps> = ({
...
               {value.length}/{MAX_DESCRIPTION_LENGTH}
             </CharacterCount>
-            <TooltipWrapper description="Save description" shortcut="Mod+Enter">
+            <TooltipWrapper
+              description="Save description"
+              shortcut={
+                navigator.platform.toUpperCase().includes("MAC")
+                  ? "⌘+Enter"
+                  : "Ctrl+Enter"
+              }
```
tyler-dane: Plz use the icons instead of just the text. Additionally, it should have a space between the elements so it looks more like this:

`⌘` + `Enter` (except instead of ⌘ it's the command icon)

This will help us give the user a consistent experience.

tyler-dane (changes requested):
Pretty close, just a few requests

HirenGajjar:
Updated — using ShortCutLabel with k="Mod" now, renders the right icon per platform.

tyler-dane (approved):

PR #1813, opened 2026-05-24
fix(web): keep now description escape local

Uarmagan:
## Summary
- Keep the Now description editor mounted and focused synchronously when `E D` opens it.
- Add regression coverage for pressing `E`, `D`, then `Escape` quickly so the global Escape shortcut does not run.

## Evidence
- Scan window started after the last automation run at `2026-05-23T14:01:13.101Z`.
- `origin/main` advanced to `3771d243b4cd7f3eb43cf741b2e66eabecacb7a1` from PR #1807, which changed Now description editing to the `E D` shortcut and added a global Escape path back to Day.
- A local regression test reproduced the bug against the current merged code: the fast `E D Escape` sequence called the global Escape handler once.
- Root cause: the description editor was opened through an async state update, so a fast following Escape could be handled before the textarea was mounted and focused.

## Validation
- `bun test --cwd packages/web src/views/Now/shortcuts/useNowShortcuts.test.tsx src/views/Now/components/TaskDescription/TaskDescription.test.tsx`
- `bun test:web`
- `bun type-check`
- `bun lint`
- `git diff --check`

### DIFF
--- a/packages/web/src/views/Now/components/TaskDescription/TaskDescription.tsx
+++ b/packages/web/src/views/Now/components/TaskDescription/TaskDescription.tsx
@@ -1,6 +1,13 @@
 import { FloppyDisk, Pencil } from "@phosphor-icons/react";
 import type React from "react";
-import { useCallback, useEffect, useRef, useState } from "react";
+import {
+  useCallback,
+  useEffect,
+  useLayoutEffect,
+  useRef,
+  useState,
+} from "react";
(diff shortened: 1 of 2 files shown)


tyler-dane:
Closing for a few reasons:
- Usage of flushSync is a smell that indicates that we haven't structured things correctly. Better to fix the root cause than add another workaround
- Tentative plan to remove the now view to make the app more simple

PR #1814, opened 2026-05-24
fix: update auth state immediately on sign out

### LINKED ISSUE #1809: Cmd Palette shows Log In and Log Out actions
### Where did this happen?

_No response_

### Expected Behavior

When logging out from the day or now view, Compass should make it obvious that the user is no longer logged in.

The cmd palette should only show either 'Log In' or 'Log out' — never both. 

### Current Behavior

After logging out from the day view, it's not clear that the user is no longer logged in
- Their calendar events are still visible
- There is no message notifying them that they successfully logged out.

Both are shown

Image

### Steps to Reproduce

1. Log in
2. Open cmd palette and select 'Log out'
3. Log out by clicking the button
4. You'll be redirected to the app again
5. Open cmd palette
6. Expected: Only 'Log In' is shown
7. Actual: 'Log In' and 'Log out' are shown

### Possible Solution

Use the user's auth state when determine which actions to display in the palette
- Ensure this is the same logic for each view. That is, the /day command palette should work the same as the /week

### Context

_No response_

jaryalarshita:
## Summary

Fixes the command palette showing both "Log In" and "Log Out" actions after signing out.

Closes #1809

## Changes Made

* Updated the `SIGN_OUT` session event handling to immediately set the authentication state to logged out instead of waiting for the async session check to complete.
* Added `handleSessionMissing()` during sign out to ensure auth-related state is cleared consistently across the app.

## Result

After logging out, the session state now updates instantly, so the command palette correctly shows only the relevant authentication actions.

### DIFF
--- a/packages/web/src/auth/compass/session/SessionProvider.tsx
+++ b/packages/web/src/auth/compass/session/SessionProvider.tsx
@@ -54,7 +54,7 @@ SuperTokens.init({
 
 export const SessionContext = createContext<CompassSession>({
   authenticated: false,
-  setAuthenticated: () => {},
+  setAuthenticated: () => { },
 });
 
 const authenticated$ = new BehaviorSubject(false);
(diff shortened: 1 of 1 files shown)


tyler-dane:
Closing, as this was fixed in a recent PR

PR #1819, opened 2026-05-28
fix(web): ignore right-click resize starts

### LINKED ISSUE #1815: Right-clicking on slider and then moving mouse results in resize
### Where did this happen?

Hosted (compasscalendar.com)

### Expected Behavior

When right-clicking on an event's slider, the context menu should open. When the user then moves their mouse (with the context menu still open), the event rectangle shouldn't change

### Current Behavior

The event rectangle changes while moving the mouse, ie the action is being treated as a resize

https://github.com/user-attachments/assets/938f7573-ea4c-4c0e-95a8-4f1fca5869db

### Steps to Reproduce

1. Create event
2. Hover over start/end of event until cursor. changes
3. Right click
4. Wait for context menu to render
5. Move mouse

### Possible Solution

_No response_

### Context

This is startling UX that will discourage ppl from using Compass

Uarmagan:
## Summary
- Prevent right-clicks from starting event resizing in the shared resize wrapper used by Day view.
- Add a regression test that verifies right-click resize starts are ignored.

Closes #1815.

## Root cause
Day view resize handles started resizing for any mouse button. Holding the right mouse button on a resize edge and moving the pointer could therefore resize the event while the user was trying to use the context menu.

## Validation
- `cd packages/web && bun test src/components/DND/Resizable.test.tsx`
- Reproduced the Day-view right-click-hold-and-move bug locally, then confirmed the event height stayed unchanged after the fix.
- `bun run type-check`
- `bun lint`

Note: `bun run test:web` is blocked by a repeatable Bun 1.2.3 crash in the unrelated `src/common/hooks/useLogoutCmdItems.test.ts` file.

### DIFF
--- a/packages/web/src/components/DND/Resizable.test.tsx
+++ b/packages/web/src/components/DND/Resizable.test.tsx
@@ -11,7 +11,7 @@ mock.module("@web/common/hooks/usePointerPosition", () => ({
 type MockResizeStart = (
   event: ReactMouseEvent<HTMLButtonElement>,
   direction: "right",
-  element: null,
+  element: HTMLElement,
 ) => void;
 
 type MockResizeStop = (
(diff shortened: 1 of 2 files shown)


Uarmagan:
Fixed in previous PR

PR #1861, opened 2026-06-09
fix(web): keep held no-op drags from opening events

### LINKED ISSUE #1785: Cursor does not transition to `move` after mouse is held down

remove334:
## Summary

- prevents held no-op drags from opening the event form
- updates the Day/Week coordinator logic for no-op drag completion
- adds a focused Day regression test

Closes #1785.

## Test plan

- `git diff --check`
- Could not run the test suite locally because `bun` is not installed in this environment.

### DIFF
--- a/packages/web/src/views/Day/interaction/DayInteractionCoordinator.test.tsx
+++ b/packages/web/src/views/Day/interaction/DayInteractionCoordinator.test.tsx
@@ -23,7 +23,15 @@ import { gridEventDefaultPosition } from "@web/common/utils/event/event.util";
 import { editEventSlice } from "@web/ducks/events/slices/event.slice";
 import { DayInteractionCoordinator } from "./DayInteractionCoordinator";
 import { dayCalendarEventRegistry } from "./registry/dayCalendarEventRegistry";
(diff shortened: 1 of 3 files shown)


tyler-dane:
Thank you for taking the time to set up Compass and submit a PR. Unfortunately, this doesn't meet the requirement for new contributors (see CONTRIBUTING.md for more details).

PR #1866, opened 2026-06-12
fix(web): stabilize day drag and drop

### LINKED ISSUE #1635: Replace `@hello-pangea/dnd` with `@dnd-kit`
No need to have two DND libs. 
Refactor the existing usage of `@hello-pangea/dnd` to use `@dnd-kit`. 

Consolidate our DND libraries, so that it's easier to maintain. We're successfully using @dnd-kit on the day view, and it's a more capable library. Convert the week view's use of DND to use dnd-kit. Ensure existing DND functionality is not broken

tyler-dane:
## Summary
- replace day task reordering with the dnd-kit sortable path
- account for live main-grid scroll position when dropping Someday events onto the calendar
- make day task drag handles stable by rendering the handle in a real in-row lane instead of a floating overlay
- remove the unnecessary task-list scrollbar gutter

closes #1635 

## Root Cause
The day task reorder affordance was visually positioned separately from the task row. After mouse reorder, handle hover and hit testing could diverge by row, leaving some task handles invisible even though keyboard reordering still worked. The Someday timed-drop path also used stale initial scroll state instead of the current grid scroll position.

## Validation
- `bunx playwright test e2e/tasks/reorder-task.spec.ts`
- `bun test --cwd packages/web src/views/Day/components/Tasks/Tasks.test.tsx`
- `bun type-check`
- `bun lint`
- `bun test:web`


tyler-dane:
TODO: fix when there are lots of tasks. We need to be able to scroll the list. 
Also consider adding a maximum number o tasks per day that are open (eg 20)

PR #1900, opened 2026-07-01
fix(web): enable keyboard navigation in recurring event scope dialog

### LINKED ISSUE #1326: Cannot select recurrence option with keyboard
### Where did this happen?

Local code with an up-to-date main branch, Hosted (app.compasscalendar.com)

### Expected Behavior

When changing a recurring event from the week view, Compass prompts the user to select between applying this change to `This Event`, `This and Following Events`, and `All Events`, in addition to cancelling the change.

When the user is presented with these options, they should be able to select the desired option using just their keyboard by typing TAB, SHIFT+TAB, SPACE, and ENTER. 

The option that is currently selected (focused) should also have different styles, so that the user can identify which option they have selected. This also applies to when the user has focused on the Cancel / Save button. 

### Current Behavior

When the user is presented with these options and attempts to use their keyboard to make a selection, nothing happens

Image

### Steps to Reproduce

1. Create a recurring event
2. Click save from the form
3. The modal will appear
4. Try typing TAB or ENTER
5. Nothing will happen

### Possible Solution

_No response_

### Context

When fixing, ensure to
- re-use existing tailwind styles for the focus color (reference the day view)
- use `floating-ui` for the modal pattern. They have a11y built-in
- Write tests to protect this a11y use-case and ensure we don't regress
- Use the `axe DevTools` Chrome extension or similar to audit your changes

Sapnilb15:
## What does this PR do?

Fixes keyboard navigation in the recurring event update scope dialog — the modal that appears when saving a recurring event. Before this change, users who rely on keyboards could not Tab, Arrow-navigate, Space, or Enter to select a scope option — every key press silently did nothing. After this fix the full standard keyboard flow works.

## Why was this PR needed?

Issue #1326 reported that pressing Tab or Enter in the dialog did nothing. Root cause: radio inputs used Tailwind classes `absolute opacity-0 w-0 h-0`. Zero-dimension elements cannot receive keyboard focus in most browsers, so every key press was silently swallowed.

Two changes fix this:

1. **Radio inputs** — Changed class from `absolute opacity-0 w-0 h-0` to `sr-only` (Tailwind's 1px clip technique). Inputs stay visually hidden but are now focusable and in the tab order. The pre-existing `peer-focus-visible:ring-accent-primary` styles on the custom radio dot now activate correctly.

2. **Modal pattern** — Replaced the hand-rolled `OverlayPanel` wrapper with floating-ui's `FloatingFocusManager` + `FloatingOverlay` + `FloatingPortal` (per the issue guidance). This provides built-in focus trapping, return-focus-on-close, Escape dismiss via `useDismiss`, and correct `role="dialog"` + `aria-modal` via `useRole`.

## What are the relevant issue numbers?

Closes #1326

## Screenshots / Recordings

After fix, full keyboard flow works:
- `Tab` — focuses first radio on dialog open, then cycles Cancel → Ok → radio group
- `↑` / `↓` — navigates between options and selects them (native radio group behavior, now unblocked by the `sr-only` fix)
- `Space` — selects the focused radio option
- `Enter` — submits with the currently selected scope
- `Escape` — closes the dialog

## Does this PR meet the acceptance criteria?

- [x] Tests added: 16 tests in `RecurringEventUpdateScopeDialog.test.tsx` covering focus, Arrow keys, Space, Enter, Escape, Tab, and mouse click flows
- [x] All tests passing: `bun test` — 16 pass, 0 fail
- [x] Follows project style guide: Biome lint clean on changed files; Conventional Commits commit messages
- [x] No breaking changes: public component API unchanged; `OverlayPanelActionButton`/`OverlayPanelActions` still reused
- [x] Reuses existing focus ring tokens (`ring-accent-primary`, `ring-offset-panel-bg`) consistent with the rest of the modal system

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -1,5 +1,6 @@
 {
   "lockfileVersion": 1,
+  "configVersion": 0,
   "workspaces": {
     "": {
       "name": "compass",
@@ -16,6 +17,7 @@
         "@babel/preset-typescript": "^7.16.0",
         "@biomejs/biome": "^2.4.12",
         "@playwright/test": "^1.60.0",
+        "@sinonjs/fake-timers": "^15.4.0",
         "@types/node": "^22.13.10",
         "@types/node-fetch": "^2.6.13",
         "jest": "^29.0.3",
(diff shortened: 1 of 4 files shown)


tyler-dane:
Thank you for taking the time to set up Compass and submit a PR. Unfortunately, this doesn't meet the requirement for new contributors (see CONTRIBUTING.md for more details).

PR #2090, opened 2026-07-13
feat(web): replace signup subscribe checkbox with post-signup release-notes prompt

tyler-dane:
## Summary

The old "Subscribe to updates" checkbox on the sign-up form was easy to miss, and it was only wired to the email/password path — users who signed up with **Continue with Google** were never offered the subscription at all.

This replaces the checkbox with a dedicated, hard-to-miss prompt shown **once, right after a new user signs up** (email/password *or* Google), asking whether they want monthly release notes, with an unsubscribe note and two clear choices:

- **Yes, Keep Me Updated** → subscribes, shows a brief "You're in!" beat, then animates into the app.
- **Nah, I don't want updates** → shows *"No problem. You can sign up using the cmd palette if you change your mind."* then animates into the app.

Both signup paths flip an in-memory `release-notes-prompt` store flag that a single globally-mounted `ReleaseNotesPrompt` (in `RootShell`) renders off of. The Google path now surfaces the backend's `createdNewRecipeUser` as `isNewUser`, so returning logins and Google-calendar reconnects don't trigger it. The subscribe API call is extracted into a shared `subscribeToReleaseNotes()` helper reused by the existing command-palette item. The prompt reuses `WelcomeModal`'s panel styling and fade/scale (reduced-motion-aware) animation.

## Simplicity

Net complexity is down: the per-form checkbox `useState` and its submit plumbing are gone, and duplicated subscribe logic is unified into one helper. The new `ReleaseNotesPrompt` carries **two `useState`** (`status` for the asking→confirmed/declined machine, `closing` for the exit animation) — both UI-only, nothing outside the component needs them. It has **no `useEffect` and no `useRef`**: `RootShell` conditionally mounts the prompt so each open is a fresh mount (no reset effect needed), and focus-on-mount uses a callback ref rather than an effect+ref pair.

## Manual Testing Steps

- [ ] Open the app and go to the sign-up form (click **Log in** → toggle to **Sign up**, or load `/day/<today>?auth=signup`). Confirm there is **no** "Subscribe to updates" checkbox — only Name, Email, Password, and Sign up.
- [ ] Complete a **new** email/password sign-up. Confirm a "Want monthly release notes?" prompt appears immediately after.
- [ ] Click **Yes, Keep Me Updated**. Confirm a brief "You're in!" confirmation shows, then the prompt fades away into the app. Open the command palette (Cmd/Ctrl+K) and confirm **"Subscribe to Updates" is gone** (proves the subscription saved).
- [ ] Sign up as another new user and click **Nah, I don't want updates**. Confirm it shows "No problem. You can sign up using the cmd palette if you change your mind." then fades away, and that **"Subscribe to Updates" is still present** in the command palette.
- [ ] Sign up with **Continue with Google** as a brand-new account; confirm the same prompt appears after the redirect. Log out and log back into that same account; confirm the prompt does **not** appear for the returning login.
- [ ] With the prompt open, press **Escape** (or click the backdrop); confirm it dismisses without subscribing.

## Test plan

- [x] `bun run type-check` — clean.
- [x] `bun run lint` (Biome) on all touched files — clean.
- [x] `bun test packages/web` affected suites — 50 pass, 0 fail (`AuthModal.test.tsx`, `google-authorization.test.ts`, `useSubscribeCmdItems.test.ts`), including a new `isNewUser`-on-signin case and the rewritten "opens the release-notes prompt after a successful sign-up" test.
- [x] Validated in real Chrome: prompt renders correctly, the "Nah" path shows the right copy and fully unmounts after dismissal, the sign-up form has no checkbox, and no console errors. The "Yes → confirmed" success state requires a live SuperTokens backend, so it was covered by unit tests rather than the local walkthrough.

### DIFF
--- a/packages/web/declaration.d.ts
+++ b/packages/web/declaration.d.ts
@@ -37,13 +37,20 @@ declare const BUILD_VERSION: string;
 interface Window {
   /** Set by Playwright prepareOAuthTestPage; disables SuperTokens session checks in e2e mode. */
   __COMPASS_E2E_TEST__?: boolean;
-  /** Semantic user-metadata bridge for e2e tests. Set by packages/web/src/auth/state/user-metadata.store.ts. */
+  /** Semantic store bridges for e2e tests. Each store sets its own key (see
+   * user-metadata.store.ts, release-notes-prompt.store.ts); keys are optional
+   * because they populate independently as their modules evaluate. */
   __COMPASS_E2E_STORE__?: {
-    userMetadata: {
+    userMetadata?: {
       getState: () => import("@web/auth/state/user-metadata.store").UserMetadataState;
       set: (metadata: import("@core/types/user.types").UserMetadata) => void;
       setLoading: () => void;
       clear: () => void;
+    };
+    releaseNotesPrompt?: {
+      getState: () => import("@web/auth/state/release-notes-prompt.store").ReleaseNotesPromptState;
+      open: () => void;
+      close: () => void;
     };
   };
   /** Session test hooks exposed by SessionProvider for e2e auth control. */
--- a/packages/web/src/auth/compass/user/util/subscribe.util.ts
+++ b/packages/web/src/auth/compass/user/util/subscribe.util.ts
@@ -1,0 +1,17 @@
+import { type UserMetadata } from "@core/types/user.types";
+import { UserApi } from "@web/api/user.api";
+import { userMetadataActions } from "@web/auth/state/user-metadata.store";
+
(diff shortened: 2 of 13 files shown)


tyler-dane:
Closing as duplicate: this feature already shipped to main in 2ef4b0dcd ("feat(web): prompt new users for release notes") while this branch was in flight. Opening a small follow-up with the two non-duplicated improvements (e2e store bridge + hook simplification) on top of main's version instead.

PR #2211, opened 2026-07-20
fix(web): fix contrast and focus-ring clipping in auth and date pickers

tyler-dane:
## Summary
- Auth modal primary button (Log in / Sign up) used `--text` for its label color against a `bg-accent` fill, which is near-invisible in both enabled and disabled states in dark-abyss; switched to `--on-accent` (the token designed for accent-filled surfaces) and raised the disabled opacity from 50% to 70% so contrast holds once dimmed.
- "Continue with Google" button repurposed the theme's `--text` token as a fake white background, which flips dark in light-beach and produces dark-on-dark text; replaced with a fixed white background so the button's contrast no longer depends on theme.
- Week grid's keyboard focus ring was painted as an inset outline directly on the scrollable, `overflow-y-hidden` section, so its top edge got clipped/painted-over by the all-day row; moved the ring to a sibling overlay `div` (toggled via `peer-focus-visible`) so it's never subject to the scroll container's own clipping.
- Sidebar month picker: zero horizontal padding on the month grid left no room for the day cells' focus ring on Sunday/Saturday, clipped by the calendar's `overflow: hidden`; added minimal padding to match. The sidebar also forced every day into a `rounded-default` rectangle via `dayClassName`, diverging from the form picker's circular days; removed that override. The sidebar's "selected day" pill separately used a `border-radius: var(--radius-default)` rectangle; changed to a circle to match. "Today" text color was duplicated across three CSS rules (base / `[data-dark]` / `[data-view="sidebar"]`) with a fragile heuristic that broke for the recurrence "Ends on" picker's event-color background; collapsed to one rule using `--accent`, which is chosen per-theme to contrast against normal surfaces.

## Simplicity
Net reduction: the "today" text-color logic went from three overlapping CSS rules (with a `[data-dark]` JS-computed heuristic) to one. No `useEffect`/`useRef`/`useState` were added or touched — every fix here is a className/CSS change.

## Manual Testing Steps
- [ ] Open the app while logged out, click "Log in" — with the email/password fields empty, confirm the "Log in" button text is clearly legible against its light-blue disabled background (not near-invisible).
- [ ] Fill in an email and password — confirm the enabled "Log in" button shows dark text on the light-blue background.
- [ ] Switch to light theme (command palette → theme, or however the app currently switches) and view the "Continue with Google" button — confirm its dark text stays readable against a white button background.
- [ ] In Week view, click into the calendar grid to give it keyboard focus (e.g. click the scrollable grid area) — confirm a continuous blue focus outline wraps the entire grid, including a visible top edge above/around the all-day event row (not cut off).
- [ ] Open the sidebar month picker, use arrow keys to move keyboard focus onto a Saturday or Sunday (rightmost/leftmost column) — confirm the focus ring renders as a full circle, not clipped at the edge.
- [ ] In the sidebar month picker, confirm the selected day and today's date render as circles (not rounded rectangles), and that today's date number is clearly visible against its background in both light and dark theme.

## Test plan
- [x] `bun run lint` — no new warnings in changed files
- [x] `bun run type-check` — passes
- [x] `bun test --cwd packages/web src/components/Sidebar/MonthPicker/MonthPicker.test.tsx` — 1 pass
- [x] `bun run test:web` — 1266 pass, 1 pre-existing unrelated failure (`handleErrorResponse > still signs the user out on an unauthorized data-endpoint response`, confirmed failing on base branch via `git stash`)
- [x] Manually verified all five fixes live via the local dev server (temp account, both themes) — contrast computed numerically (disabled button ~4.9:1, up from ~1.2:1) and focus rings/circles confirmed via DOM inspection + screenshots

### DIFF
--- a/.claude/skills/simplify/SKILL.md
+++ b/.claude/skills/simplify/SKILL.md
@@ -157,7 +157,7 @@ state that nothing else needs), keep it — but say why in the commit message
 ## Repo-Specific Conventions
 
 - Use path aliases, not deep relative imports: `@compass/core`,
-  `@compass/backend`, `@compass/scripts`, `@web/*`, `@core/*`
+  `@compass/backend`, `@compass/scripts`, `@compass/sync`, `@web/*`, `@core/*`
 - Shared web/backend contracts belong in `packages/core` and use Zod — match
   the existing import style in the file you're editing
 - Zustand stores follow store + module actions + plain selectors (see
@@ -230,6 +230,7 @@ Match the check to the packages actually touched — don't default to a broad
 suite:
 
 - `packages/core` changes → `bun run test:core`
+- `packages/sync` changes → `bun run test:sync`
 - `packages/web` changes → `bun run test:web`
 - `packages/backend` changes → `bun run test:backend`
 - `packages/scripts` changes → `bun run test:scripts`
--- a/.github/workflows/test-unit.yml
+++ b/.github/workflows/test-unit.yml
@@ -89,7 +89,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        project: [core, web, backend, scripts]
+        project: [core, sync, web, backend, scripts]
 
     steps:
       - name: Check out Git repository
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -23,7 +23,9 @@ cp compass.example.yaml compass.yaml
 bun install
 bun dev:web
 bun dev:backend
+bun dev:sync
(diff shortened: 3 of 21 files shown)


tyler-dane:
Superseded by a clean branch rebased off current main — the original branch had unrelated merge conflicts from in-progress packages/sync scaffolding that predated this change. Same fix, same diff, opened as a new PR to avoid resolving conflicts in code outside this change's scope.

PR #2230, opened 2026-07-21
fix(web): correct low-contrast text across light and dark themes

tyler-dane:
## Summary

Several surfaces paired text against a background with insufficient contrast in one or both themes:

- The recurrence mini-calendar's weekday header (`S M T W T F S`) used `--text-muted`, a token only calibrated against the app's fixed near-black/cream surfaces, against a dynamic event-colored popup background — landing around 1.9:1. Fixed to match the identical, already-correct sibling rule used for the day numbers.
- The recurrence "Ends on" date input never had a text `color` set, so it fell back to a browser default that was invisible against its background in one theme. Now uses the existing `theme.getContrastText()` helper.
- The Save button's background was darkened into contrast's "mid-tone dead zone," leaving its text at a marginal ~5:1. Removed the extra darken step; now 6.9–8.8:1 in both themes.
- Several buttons used `bg-accent`/`bg-accent-secondary` with `text-text` instead of `text-on-accent` (MobileGate, AuthButton's primary CTA, Cleanup, BackendDown, NotFound, and two toast components) — in one theme both colors are dark, making the label nearly invisible.
- The recurrence frequency dropdown (`Day`/`Week`/`Month`/`Year`) had a hardcoded dark-theme-only hex for its closed-state label (no override at all — fell back to react-select's default `#333`) and its open-menu options used another hardcoded hex instead of the theme-aware contrast helper.
- Two inputs (delete-account confirmation, command palette search) had their typed-text and placeholder colors inverted from every other input in the app.

All fixes reuse the app's existing token system (`--on-accent`, `--text`, `theme.getContrastText()`) — no new colors were introduced.

## Simplicity

Net reduction: `FreqSelect.tsx`'s `option` style function computed the same `isSelected ? bgBright : isFocused ? bgDark : ...` background twice (once for the new contrast calculation, once for the existing `backgroundColor` field) — the two were consolidated into one `optionBg` value. No `useEffect`, `useRef`, or `useState` was added, removed, or touched by this change.

## Manual Testing Steps

- [ ] Open the app, start a session, and open an existing timed event's edit form. Confirm the "Save" button's label is clearly legible (dark text on a light blue button).
- [ ] In that same form, click "Repeat" to enable recurrence. Confirm the "Day" frequency dropdown's label is clearly legible, then open the dropdown and confirm all four options (Day/Week/Month/Year) have legible text against their backgrounds, including on hover.
- [ ] Click the "Ends on" date field. Confirm the weekday header row (S M T W T F S) in the popup calendar is clearly legible, then pick a date and confirm the typed date text in the field is clearly legible.
- [ ] Visit the app with `?auth=login` appended to the URL (or otherwise open the login modal). Confirm the "Log in" button's text is clearly legible.
- [ ] Switch the app to the light theme (via local storage key `compass.theme` = `light-beach`, or the in-app theme switcher if available) and repeat the above checks — all should remain legible.
- [ ] On a mobile device or with a mobile user agent, confirm the "Join Mobile Waitlist" button text is clearly legible against its dark button.

## Test plan

- [x] `bun run lint` — passes (5 pre-existing warnings, none in touched files)
- [x] `bunx typescript@7.0.2 -p packages/web/tsconfig.app.json --noEmit` — passes with no errors
- [x] `bun run test:web` — 1266 passed, 1 failed (`handleErrorResponse > still signs the user out on an unauthorized data-endpoint response`); confirmed this failure pre-exists on this branch before these changes and is unrelated to the diff
- [x] Manually verified all fixes render correctly in Chrome (both the isolated preview browser and a real Chrome profile), checked console/network for errors — none found

### DIFF
--- a/.claude/skills/simplify/SKILL.md
+++ b/.claude/skills/simplify/SKILL.md
@@ -157,7 +157,7 @@ state that nothing else needs), keep it — but say why in the commit message
 ## Repo-Specific Conventions
 
 - Use path aliases, not deep relative imports: `@compass/core`,
-  `@compass/backend`, `@compass/scripts`, `@web/*`, `@core/*`
+  `@compass/backend`, `@compass/scripts`, `@compass/sync`, `@web/*`, `@core/*`
 - Shared web/backend contracts belong in `packages/core` and use Zod — match
   the existing import style in the file you're editing
 - Zustand stores follow store + module actions + plain selectors (see
@@ -230,6 +230,7 @@ Match the check to the packages actually touched — don't default to a broad
 suite:
 
 - `packages/core` changes → `bun run test:core`
+- `packages/sync` changes → `bun run test:sync`
 - `packages/web` changes → `bun run test:web`
 - `packages/backend` changes → `bun run test:backend`
 - `packages/scripts` changes → `bun run test:scripts`
--- a/.github/workflows/test-unit.yml
+++ b/.github/workflows/test-unit.yml
@@ -89,7 +89,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        project: [core, web, backend, scripts]
+        project: [core, sync, web, backend, scripts]
 
     steps:
       - name: Check out Git repository
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -23,7 +23,9 @@ cp compass.example.yaml compass.yaml
 bun install
 bun dev:web
 bun dev:backend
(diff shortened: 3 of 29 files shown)


tyler-dane:
Replaced by #2231, which cherry-picks only this PR's two web commits onto current `main`. A normal merge here produces unrelated sync/core conflicts because those earlier changes landed on `main` under different commit IDs.

PR #2283, opened 2026-07-22
feat(sync): non-destructive calendar repair into a fresh generation

tyler-dane:
## What

Adds a non-destructive **calendar repair**: when a provider calendar's stored
state is untrustworthy (invalid sync cursor, inconsistent local rows), rebuild
it from scratch into a fresh occurrence *generation* and activate that
generation atomically, so reads never see a half-built calendar.

## How it works

- **Build alongside, then flip.** The old generation stays the one reads serve
  for the entire rebuild. A full unwindowed provider pass repopulates a new
  generation via the shared `ProviderPageApplier`. Only once the pass yields a
  durable cursor do we activate the new generation, advance the cursor, and
  clean up the old rows — a single-field flip (`activeGeneration`).
- **Fail safe.** If the pass finishes without a cursor, we return `incomplete`
  and leave the old generation active and untouched. Nothing is destroyed on a
  failed repair.
- **Crash-safe / resumable.** The new generation is chosen so a retry resumes
  the same in-flight one rather than stranding a partial generation each attempt.
  Every write is an idempotent provider-identity upsert, and activation, cursor
  advance, and cleanup each run after the pass and are individually idempotent.

## Two stores, two treatments (read before flagging)

Occurrences are **generation-isolated** — the unique index includes
`generation`, because the range read serves occurrences by generation and the
old and new sets must coexist during a repair.

Events are **generation-watermarked**, not isolated: no user read path filters
events by generation, so exactly one canonical doc exists per provider identity
and a repair bumps its generation *in place*. That in-place bump is precisely
the signal stale-detection uses — genuinely-deleted events strand below the new
generation and get cleaned up, while re-seen ones ride forward. Adding
generation to the events unique index would collide the second import and defeat
that signal. This asymmetry is documented at `upsertByProviderIdentity`.

## Tests

`calendar-repair.service.test.ts` covers: rebuild + activate + clear old,
`incomplete` leaves the old generation intact, a Compass-owned (unlinked) event
survives, an in-flight generation resumes instead of bumping again, and
convergence when run twice. Occurrence-store isolation and the index
reconciliation have their own repository/manifest tests.

`bun run test:sync` — 469 pass, 0 fail.

### DIFF
--- a/packages/sync/src/domain/calendar-import.service.ts
+++ b/packages/sync/src/domain/calendar-import.service.ts
@@ -152,11 +152,14 @@ class ImportRun {
     private readonly resource: SyncResourceRecord,
     now: () => Date,
   ) {
+    // Import writes the active (live) generation — the one reads serve. Only a
+    // repair stages a separate importGeneration; the first import runs with
+    // active and import both at 0, so it populates the generation reads serve.
     this.#applier = new ProviderPageApplier(
       deps.events,
       deps.occurrences,
       calendar,
-      resource.importGeneration,
+      resource.activeGeneration,
       now,
     );
   }
--- a/packages/sync/src/domain/calendar-pull.service.test.ts
+++ b/packages/sync/src/domain/calendar-pull.service.test.ts
@@ -228,6 +228,35 @@ describe("pullCalendarChanges", () => {
     expect(result.changed).toBe(1);
   });
 
+  it("writes into the active generation when a prior repair left one staged", async () => {
+    // A repair bumped importGeneration ahead of activeGeneration but never
+    // activated (crashed/incomplete). A pull must still write the live (active)
+    // generation, or its new events land in a generation reads never serve.
+    const calendar = await seedCalendar();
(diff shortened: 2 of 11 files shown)


tyler-dane:
Adversarial review resolved:
- **Confirmed safe** by review: the events-store generation *watermark* (single doc per provider identity, bumped in place) — no live read path filters events by generation; only occurrence reads are generation-gated and they flip atomically. Owner-scoping verified on every write/delete.
- **Fixed (was Critical):** import and pull now write the *active* generation, not `importGeneration`. Previously an incomplete repair (which bumps `importGeneration` ahead of active) would have stranded a later pull's occurrences in a generation reads ignore. Regression test added in `calendar-pull.service.test.ts`.
- **Documented (operational):** the startup index reconciliation's inline drop-and-rebuild is unsafe once a collection carries real data; noted inline that a populated-collection key change needs a rolling migration. Sync serves no production data yet.

PR #2475, opened 2026-07-30
chore(sync): unmount legacy SyncRoutes and drop its startup webhook warning

tyler-dane:
## Summary

First standalone-safe step toward deleting `packages/backend/src/sync/` wholesale (the legacy in-backend Google-sync engine, now fully superseded by the standalone Sync microservice everywhere).

- `servers/express/express.server.ts` no longer mounts `SyncRoutes` — the legacy engine's own Express routes (Google webhook/import endpoints) were still wired in but unreachable in practice on any Sync-delegated deployment (prod and self-host both are, as of the last two cutovers).
- `app.ts` no longer calls `warnIfWebhookNotPublicHttps` at startup — a sanity check specific to the legacy engine's own webhook config, meaningless once nothing routes to it.

`packages/backend/src/sync/` itself is untouched — still compiles standalone, still has its own test coverage, nothing inside it changed. This just removes the two places outside `sync/` that reference it for routing/startup, clearing the way for later PRs to finish decoupling test infrastructure before the final wholesale deletion.

## Test plan
- [x] `bunx tsc --noEmit` clean
- [x] `./node_modules/.bin/biome check` clean
- [x] `bun run test:backend`: 54 failures, exact zero-delta match against the established pre-existing local-sandbox baseline
- [ ] CI green

### DIFF
--- a/packages/backend/src/app.ts
+++ b/packages/backend/src/app.ts
@@ -1,7 +1,6 @@
 import { CONFIG } from "@backend/common/constants/config.constants";
 import mongoService from "@backend/common/services/mongo.service";
 import { initExpressServer } from "@backend/servers/express/express.server";
-import { warnIfWebhookNotPublicHttps } from "@backend/sync/services/watch/google-watch-config";
 import { logger } from "./init"; //must be first import
 import { stopPostHogLogs } from "./logging/posthog-logs";
 import { createServer, type Server } from "node:http";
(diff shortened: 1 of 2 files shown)


tyler-dane:
Closing without merging: unmounting SyncRoutes prematurely broke sync.controller.db.test.ts's real supertest integration tests (they hit the actual mounted routes and now 404). Those tests live inside packages/backend/src/sync/, which itself isn't being deleted yet — this ordering assumption was wrong. Folding this change into the final wholesale sync/ deletion PR instead, where sync.controller.db.test.ts is deleted in the same commit.

PR #2529, opened 2026-08-01
fix(web): keep the signed-out message alive and unqueued

tyler-dane:
## Problem

Investigating a report of "my event disappeared and I got no indication of an error", two independent ways the one message telling a user their write failed could never reach them:

1. **A document navigation destroyed it.** `signOut()` raises a toast and then calls `window.location.assign("/week")`. That is a full page load, so the toast is torn down before it can be read. The early return means this only happens when the user is *not* already on the calendar route — which is precisely the case where the message matters, since they are also being moved away from where they were.

2. **The toast could be silently queued out.** `ToastContainer` had `limit={1}`, so any routine toast already on screen (an undo notice, an autosave confirmation) pushes a critical "you've been signed out" into a queue the user may never see.

Together: the user's write fails, their optimistic edit rolls back, and nothing explains why.

## Changes

- `signOut` routes in-app via the router singleton instead of assigning `window.location`. `SessionExpiredToast` already reaches the router this way, including the dynamic import that avoids the module cycle, so this follows an established path rather than inventing one.
- `limit={1}` → `limit={3}`.

## Tests

The navigating branch of `signOut` had no coverage — the existing test pins the *already-on-calendar* case and notes it does so specifically to skip the jsdom-unsupported navigation. Adds a test for the branch that was actually broken.

24/24 api.util tests pass, 1519/1519 web tests pass (checked for `mock.module` leakage across files), type-check and lint clean.

## Not included

The mutation layer still swallows 401 in `handleError`, deferring to this interceptor. That is now a safe assumption again since the interceptor's message survives, but it is worth revisiting alongside a `MutationCache.onError` backstop.

### DIFF
--- a/packages/web/src/api/util/api.util.test.ts
+++ b/packages/web/src/api/util/api.util.test.ts
@@ -268,4 +268,28 @@ describe("handleErrorResponse", () => {
     expect(signOutSpy).toHaveBeenCalledTimes(1);
     signOutSpy.mockRestore();
   });
+
+  it("leaves a non-calendar route through the router, not a document navigation", async () => {
+    // A document navigation tears down the toast signOut just raised, so on
+    // every route that actually needed it the "you've been signed out" message
+    // was destroyed before it could be read. Routing in-app keeps it alive.
+    window.history.pushState({}, "", "/day/2026-07-31");
+    const navigate = mock((_options: { to: string }) => Promise.resolve());
+    mock.module("@web/routers", () => ({ router: { navigate } }));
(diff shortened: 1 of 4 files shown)


tyler-dane:
Note: the `ToastContainer` limit raise (1 → 3) was reverted from this PR. It broke `calendar-experience` "supports keyboard selection" consistently across all 3 Playwright retries — additional simultaneous toasts appear to overlay or steal focus from the calendar combobox. e2e passed once isolated.

The underlying gap is real (a critical toast can still be silently queued behind a routine one), but it needs that interaction understood before it lands. Tracked as a follow-up.

PR #2601, opened 2026-08-05
Command palette: fuzzy search, About action, fix Settings shortcut

tyler-dane:
## Summary

Audited the command palette against the goal of being the "go-to place whenever the user has something they want to do but aren't sure how to do it yet." Findings and fixes:

- **Rigid search**: the palette only matched an exact case-insensitive substring of an item's *label*. Typing "day page" or "bug" found nothing. Replaced it with a small hand-rolled fuzzy matcher (`command-palette.search.ts`) that scores label + a new per-item `keywords` synonym list — tiered exact/prefix/word-prefix/substring/subsequence matching, multi-word "AND" queries, ranked within each section (section grouping/order is unchanged; it's structural to the UI). Added keywords across every existing item so synonyms like "day view"/"day page"/"daily" all reach `Go to Day`, and "bug"/"report"/"issue" reach `Share Feedback` — no new "report bug" action was re-added (a prior one was deliberately removed; tests still assert its absence).
- **Wrong shortcut hint**: `Show accounts` displayed the `Mod+,` keycap, but that shortcut opens the whole Settings modal (of which accounts is one section). Renamed the item to `Settings` (kept the shortcut, now accurate) with `accounts`/`preferences` as search keywords so it's still found by "accounts".
- **Version row was a dead end**: `Version: x.y.z` only copied text to the clipboard. Replaced it with an `About Compass` action that opens a small modal (version + copy button + social links). Extracted the social links (X, LinkedIn, GitHub) that were previously hard-coded only in the WelcomeModal footer into a shared `social.constants.ts`, and reused them in both places.

No dependency was added (fuzzy matcher is hand-rolled, ~90 lines including comments); no registry rewrite; contextual visibility (auth/view/state-gated items via hooks) is unchanged.

## Simplicity

- The matcher is a single pure module (`command-palette.search.ts`) with two small exported functions; no new abstraction layer over the existing per-item-hook pattern.
- Kept the palette's structural section grouping — ranking happens *within* each section rather than introducing a global re-bucketing scheme, which would have required more invasive changes to the floating-ui list-navigation wiring.
- Reused the existing `OverlayPanel` component for the new About modal instead of building bespoke modal chrome, matching `SettingsModal`.
- One shared `social.constants.ts` file replaces two copies of the same three hard-coded URLs.

## Automated validation

- `bun run lint` — clean
- `bunx typescript@7.0.2 --noEmit`, `-p packages/web/tsconfig.app.json --noEmit`, `-p packages/web/tsconfig.test.json --noEmit` — clean
- `bun run knip` — clean
- `bun run test:web` — full suite passes except 15 pre-existing `LifeView` failures unrelated to this change (verified via `git stash` against the base branch: same 15 tests fail identically before this diff, caused by a `QueryClientProvider` setup issue when that spec runs in isolation from other spec files in its worker group)

## Independent review

Self-reviewed by re-reading the full diff and hand-verifying the matcher's scoring against every existing/new test case (tie-breaks, tier ordering, subsequence floor) before running the suite. No separate reviewer agent pass was run on this change.

## Test plan

- `bun run lint`
- `bunx typescript@7.0.2 --noEmit`
- `bunx typescript@7.0.2 -p packages/web/tsconfig.app.json --noEmit`
- `bunx typescript@7.0.2 -p packages/web/tsconfig.test.json --noEmit`
- `bun run knip`
- `bun run test:web`
- New tests: `command-palette.search.test.ts` (matcher unit tests — synonyms, AND semantics, tie-breaks, subsequence floor), `About/AboutModal.test.tsx` (open/close, version + copy, social links)
- Updated tests: `CommandPalette.test.tsx` (fuzzy-filter behavior, About row), `more.cmd.constants.test.ts` (About item), `useShowAccountsCmdItems.test.ts` (rename)

---
_Generated by [Claude Code](https://claude.ai/code/session_014UMSmLvajMQzDiR3p9HmtU)_


tyler-dane:
The `unit (web)` failure on this PR (run [31039143581](https://github.com/SwitchbackTech/compass-calendar/actions/runs/31039143581)) is 15 `LifeView` test failures, all `Error: No QueryClient set, use QueryClientProvider to set one` from `SidebarStatusBar`. This is pre-existing and unrelated to this diff:

- Locally, `git stash`-ing every change in this PR and re-running the exact same suite against the unmodified base commit (`5bd5dd0`) reproduces the identical 15 failing test names.
- Scoping a run to only `packages/web/src/views/Life/` (isolated from its sibling spec files) reproduces the same `QueryClientProvider` error on both the base commit and this branch — it's a test-isolation issue in how that spec's provider tree is set up when it doesn't run alongside its normal file group, not something touched by this change.
- The command-palette-specific tests (including the new/updated ones in this PR) all pass in the full suite; the only failures are the 15 pre-existing `LifeView` ones.

Not fixing this here since it's outside the scope of the command palette audit — flagging in case a separate fix is wanted.

---
_Generated by [Claude Code](https://claude.ai/code)_

tyler-dane:
Update: the rerun of the failed `unit (web)` job reproduced the identical 15 `LifeView` failures — so this is deterministic for that specific `push`-triggered workflow run, not a one-off flake. It's tied to that run specifically: the separate `pull_request`-triggered run of this same commit (which is the one that gates this PR) passed `unit (web)` cleanly, same as every other check. Combined with the earlier `git stash` reproduction against the unmodified base commit, this confirms it's a pre-existing test-infrastructure issue unrelated to this diff, not something a rerun will resolve.

Not chasing further — out of scope for the command palette audit. The PR's actual gating checks are green.

---
_Generated by [Claude Code](https://claude.ai/code)_

tyler-dane:
Root-caused and fixed the `LifeView` failures in 23554c8 — this was a real bug, not just a flake I should shrug off.

**Cause:** `LifeView.test.tsx` rendered `<LifeView />` with a bare `render()`, no `QueryClientProvider`. `LifeView` mounts the shared `SidebarShell` → `SidebarStatusBar`, which calls `useHasPendingEventMutations` → `useMutationState` → `useQueryClient()`, which throws without a provider. The web test suite runs as a single sequential Bun process (not `--parallel`/isolated — see the comment in `test-parallel.ts`), so whether this crashed depended on incidental state left behind by whichever file happened to run before it in that process — explaining why it passed on one CI run and failed on another with the identical commit.

**Fix:** wrap the test's render in a real `QueryClientProvider` (matching `MainGrid.test.tsx`'s pattern), which also revealed a second latent issue: `SidebarStatusBar` renders its own (usually empty) `role="status"` node, so once it mounts for real, the existing `screen.getByRole("status")` assertions became ambiguous against the life-summary status text. Disambiguated with a small helper that matches on the summary's content.

Verified: full `bun run test:web` is 1694/1694 passing locally, plus lint/type-check/knip clean. Waiting on CI to confirm, then merging.

---
_Generated by [Claude Code](https://claude.ai/code)_

tyler-dane:
Turns out someone beat me to it: `main` already had PR #2600 merged, which fixes the exact same `LifeView`/`QueryClientProvider` bug I diagnosed above (same root cause — `SidebarStatusBar` landed via #2598 and needed a provider the test never gave it), via `renderWithStore` instead of my manual wrap. Merged `main` into this branch (a9865b5) and resolved the conflict by taking main's version of `LifeView.test.tsx`, dropping my now-redundant duplicate fix.

Verified post-merge: lint clean, all three type-check configs clean, knip clean, full `bun run test:web` is 1714/1714 passing. Waiting on CI for the merge commit, then merging.

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #2606, opened 2026-08-05
Remove @tanstack/react-router vendor patch (upgraded to 1.170.20)

tyler-dane:
## Summary
Upstream TanStack Router v1.170.19+ ships a fix for the race condition during cold `beforeLoad` redirects (PR #7805). Removes the bun vendor patch and updates the router dependency.

- Cold `/` → `/week` redirect now works cleanly
- Verified: no "Unhandled falsy error" crash on cold load
- Removes 60-line patch file and patchedDependencies entry

## Test Plan
- [x] Type-check passes
- [x] Cold load `/` redirects to `/week` without errors
- [x] App renders calendar normally
- [x] No console errors related to redirect

🤖 Generated with Claude Code


tyler-dane:
Revert this upgrade: 1.170.20 introduces a 2-day-old loading-pipeline rewrite that breaks app behavior. AuthModal URL-param tests fail in isolation; full suite shows 623 fails/49 errors/412s vs main's 30s.

Upstream PR #7805 is confirmed to fix the race, but needs stabilization. Retry in ~3 weeks when TanStack has had no release for 1+ week and AuthModal.test.tsx passes in isolation on the new version.

The existing patch on 1.170.17 is safe and self-documenting. Keep it.

PR #2619, opened 2026-08-07
fix(web): remove border from UpNextCard to reduce visual clutter

tyler-dane:
Removes the full border from the UpNextCard component, keeping only the visual line above the sidebar action buttons for a cleaner appearance.

### DIFF
--- a/.cursor/settings.json
+++ b/.cursor/settings.json
@@ -1,0 +1,7 @@
+{
+  "plugins": {
+    "mongodb": {
+      "enabled": true
+    }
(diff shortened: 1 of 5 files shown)


tyler-dane:
Superseded: this fix already landed on main as #2620 (fce0ff405). This branch was cut before #2608/#2613, so merging it now would revert the UpNextCard "Now" state, the "All clear" empty state, and the conference Join link. Closing rather than rebasing since there's nothing left to carry over.

PR #2625, opened 2026-08-07
fix(web): eliminate flash when creating and saving a new event

tyler-dane:
## Summary

When a user creates an event on the grid and saves it, the optimistic write and draft teardown were happening in separate React commits, causing a one-frame gap where the grid rendered neither the draft nor the saved card.

**Solution:** Thread an `onOptimisticApplied` callback through `create()` that fires synchronously inside `onMutate`, before the mutation executes. This ensures the cache write and draft discard happen in the same task, so React batches them into a single commit.

## Changes

- `useEventMutations.ts`: add optional `onOptimisticApplied` callback to `CreateVariables`, invoked in `buildMutation`'s `onMutate` immediately after the optimistic cache write
- `useSaveEventForm.ts`: pass `closeEventForm` as the callback when creating, instead of calling it after `mutate()` returns
- `useSaveEventForm.test.tsx`: add test to verify create initiates a mutation with the callback mechanism

This fix applies to Week/Day timed events and all-day events automatically—no separate handling needed.

## Test plan

- [x] \`bun run test:web packages/web/src/views/Forms/hooks/useSaveEventForm.test.tsx\` — 2 pass
- [x] \`bun run test:web packages/web/src/views/Week/components/Grid/MainGrid/MainGrid.test.tsx\` — 17 pass
- [x] \`bun run type-check\` — all pass
- [x] \`bun lint\` — no errors

🤖 Generated with Claude Code

### DIFF
--- a/.cursor/settings.json
+++ b/.cursor/settings.json
@@ -1,0 +1,7 @@
+{
+  "plugins": {
+    "mongodb": {
+      "enabled": true
+    }
+  }
+}
--- a/bun.lock
+++ b/bun.lock
@@ -8,6 +8,7 @@
         "@compass/backend": "*",
         "@compass/core": "*",
         "module-alias": "^2.2.2",
+        "posthog-js": "^1.409.0",
         "tslib": "^2.4.0",
       },
       "devDependencies": {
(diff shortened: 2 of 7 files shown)


tyler-dane:
Superseded: this fix already landed on main as #2626 (b3d1b1f8e) — the onOptimisticApplied batching and its explanatory comment are on main today. This branch was cut before #2607/#2616/#2633, so merging it now would revert the recurrence-scope promotion feature, the undo/redo mutation callbacks, and the editableContent wire-boundary strip. Closing rather than rebasing since there's nothing left to carry over.

PR #2631, opened 2026-08-07
fix(web): point logout teardown at the real sse and sync-state modules

tyler-dane:
Unblocks `type-check` on main.

`logout.teardown.ts` (from #2627) imports two things that do not exist:

- `sse` from `@web/events/sse/sse.client` — that directory does not exist; the SSE client lives at `@web/sse/client/sse.client` and exports `closeStream` as a named function, not an `sse` namespace object
- `clearGoogleSyncIndicatorOverride` from `@web/auth/google/util/google.auth.util` — not exported there; it lives in `@web/auth/google/state/google.sync.state`

Its test mocked those same two phantom paths, so the mocks intercepted nothing and the suite failed.

## Verification

- `bun run type-check` passes (was 2 errors)
- `bun test ./src/auth/compass/session/logout.teardown.test.ts` — 3 pass, 0 fail

Part of restoring main, which has been red since #2617 broke `bun install --frozen-lockfile` and 11 PRs merged without CI signal.

### DIFF
--- a/packages/web/src/auth/compass/hooks/useLogout.test.ts
+++ b/packages/web/src/auth/compass/hooks/useLogout.test.ts
@@ -17,15 +17,27 @@ const mockUseSession = mock();
 const clearAccountScopedClientState = mock();
 let signOut = spyOn(session, "signOut");
 
+// bun's mock.module is global and leaks into every other test file in the
+// run. Spread the real module so unrelated suites still see its full surface;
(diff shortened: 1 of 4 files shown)


tyler-dane:
Superseded by #2633, which carried these three commits (cherry-picked) plus the rest of the fixes needed to get every check green in one PR. Merged as 9b0df6d4b.

PR #2668, opened 2026-08-07
fix(web): dismiss shortcuts overlay through app shortcuts

tyler-dane:
## Summary
- Route the live shortcuts overlay Escape path through `useAppShortcut` (`ignoreAppLock` + `ignoreInputs: false`) so clear-search / dismiss honors the app-lock policy instead of a hand-rolled document listener.
- Conditionally render the shortcuts dialog only while open (enter via `starting:-translate-x-full`) instead of a permanently mounted `aria-hidden` dialog with compensating `tabIndex`s.
- Keep stacking at `ZIndex.MAX` (no bump to `Z_INDEX_MODAL`).

## Simplification
- Collapsed redundant filter paths, moved filter work after the open-gate, and replaced `useRef` + `setTimeout` focus with a stable callback ref.
- Separate commit: `refactor(web): simplify shortcuts overlay open-state wiring`.

## Test plan
- [x] `bun run type-check`
- [x] `bun test:web` (1820/1820) + focused ShortcutsOverlay / useSidebarShortcuts tests
- [x] `bun run lint` (pre-existing warnings only)
- [x] Functional Playwright e2e (15/15 non-axe + focus-visible); axe color-contrast failures observed locally appear unrelated (datepicker outside-month `#68747e` on `#121a23`) — confirm CI
- [x] Independent code review: clean
- [x] Manual browser smoke on http://localhost:9083: `Shift+/` opens overlay + sets `data-app-locked` + focuses search; first Escape clears search; second Escape closes and clears lock

### DIFF
--- a/packages/web/src/components/Sidebar/ShortcutsOverlay/ShortcutsOverlay.test.tsx
+++ b/packages/web/src/components/Sidebar/ShortcutsOverlay/ShortcutsOverlay.test.tsx
@@ -50,9 +50,9 @@ describe("ShortcutsOverlay", () => {
       wrapper,
     });
 
-    const overlay = screen.getByRole("dialog", { name: "Keyboard shortcuts" });
-
-    expect(overlay.firstElementChild?.className).toContain("translate-x-0");
+    expect(
+      screen.getByRole("dialog", { name: "Keyboard shortcuts" }),
+    ).toBeInTheDocument();
     expect(screen.getByText("Shortcuts")).toBeInTheDocument();
     expect(
       screen.getByText("Keyboard shortcuts for Day view"),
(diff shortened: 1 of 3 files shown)


tyler-dane:
## Follow-up after CI e2e

First CI run failed on axe `color-contrast` for sidebar datepicker outside-month days and (locally) event-form controls. Root cause: switching the shortcuts overlay to **conditional unmount** removed a permanently mounted covering dialog that had been papering over those checks.

Adjustment in `d1126c5f8`:
- Prefer F3's other option — `inert` when closed — so Escape/`useAppShortcut` cleanup still ships and the slide animation stays.
- Drop outside-month `opacity: 0.8` (was blending `--text-muted` to ~3.66:1 on `--surface-panel`).
- Broader sidebar form contrast debt (timepicker `#333` defaults, `text-muted` on `surface-overlay`) remains real and should be a dedicated follow-up; not bundled here.

PR #2789, opened 2026-08-17
fix(sync): import selected hidden calendars and offset-only events

tyler-dane:
## Summary

Compass could report a healthy Google connection while some events never appeared on the grid. Two import bugs caused that:

1. Calendar discovery treated Google `hidden: true` as inactive even when `selected: true`, so those calendars were never imported, never listed in the sidebar, and never counted in `lastSyncedAt`.
2. Timed events with an RFC3339 offset but no IANA `timeZone` were dropped as `unmappableSchedule`. Google allows that shape; the reader then incremented `skipped` with no log.

This keeps selected-but-hidden calendars active, imports offset-only timed events (instant preserved, zone UTC), and logs skip reason plus provider event id (never title).

Existing calendar-list rediscovery enqueues `initialImport` when a calendar becomes active, so already-connected accounts pick this up on the next list pass.

## Simplicity

The discovery rule is a one-line superset of the previous check. Offset-only events reuse the existing GMT-offset UTC fallback instead of adding a new schedule path.

## Automated validation

- Focused Google adapter/normalizer/reader tests: 63 pass
- `bun run test:sync:fast`: 328 pass, 0 fail
- `bun run lint`: exit 0 (pre-existing warnings only; none in these files)

## Independent review

Not run.

## Test plan

- `bun test packages/sync/src/providers/google/google-calendar.adapter.test.ts packages/sync/src/providers/google/google-event.normalizer.test.ts packages/sync/src/providers/google/google-event-reader.adapter.test.ts`
- `bun run test:sync:fast`
- `bun run lint`

After deploy, Refresh or wait for calendar-list rediscovery so selected-but-hidden calendars import. Confirm the missing Walk/Haircut events and that those calendars appear in the Compass sidebar.

<div><a href="https://cursor.com/agents/bc-e98d49a5-de5e-4bfd-a225-b349aaba3c85?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-e98d49a5-de5e-4bfd-a225-b349aaba3c85&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>

### DIFF
--- a/packages/sync/src/providers/google/google-calendar.adapter.test.ts
+++ b/packages/sync/src/providers/google/google-calendar.adapter.test.ts
@@ -277,6 +277,39 @@ describe("GoogleCalendarAdapter", () => {
     expect(active).toEqual({ live: true, gone: false, hidden: false });
   });
 
+  it("keeps a hidden calendar active when Google is still showing it", async () => {
+    const api = new FakeCalendarListApi([
+      page({
+        items: [
+          entry({ id: "shown-hidden", hidden: true, selected: true }),
+          entry({ id: "unselected-hidden", hidden: true, selected: false }),
+          entry({ id: "hidden-omitted-selected", hidden: true }),
+          entry({
+            id: "deleted-but-selected",
+            deleted: true,
+            selected: true,
+          }),
+        ],
+        nextSyncToken: "s",
+      }),
+    ]);
+    const { adapter } = adapterWith(api);
+
+    const { calendars } = await adapter.discoverCalendars({
+      accessToken: "at",
+    });
+    const active = Object.fromEntries(
+      calendars.map((c) => [c.providerCalendarId, c.active]),
+    );
+
+    expect(active).toEqual({
+      "shown-hidden": true,
+      "unselected-hidden": false,
+      "hidden-omitted-selected": false,
+      "deleted-but-selected": false,
+    });
+  });
+
   it("falls back through summaryOverride, summary, then id for the display name", async () => {
     const api = new FakeCalendarListApi([
       page({
--- a/packages/sync/src/providers/google/google-calendar.adapter.ts
+++ b/packages/sync/src/providers/google/google-calendar.adapter.ts
@@ -221,9 +221,12 @@ async function mapCalendar(
     color: item.backgroundColor || null,
     eventLabels: await api.getEventLabels(item.id),
     primary: item.primary === true,
-    // deleted appears in incremental results; hidden means the user removed it
-    // from their list. Either makes the calendar inactive, not gone.
-    active: item.deleted !== true && item.hidden !== true,
+    // Deleted calendars are gone. Hidden-from-list is not the same as
+    // "Google is not showing this": the Calendar UI still displays events
+    // when selected is true. Import those so connection health cannot stay
+    // green while a selected calendar is silently skipped.
+    active:
+      item.deleted !== true && (item.hidden !== true || item.selected === true),
     accessRole,
     capabilities: CAPABILITIES_BY_ROLE[accessRole],
   };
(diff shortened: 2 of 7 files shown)


tyler-dane:
Reviewed this while investigating the same report. The normalizer fix and the reader logging look right to me — logging at the reader is nice, since it covers the pull path too, which never read `page.skipped` at all.

Two things worth a second look.

**1. The `hidden`/`selected` discovery change conflicts with a documented acceptance scenario.**

`docs/acceptance/google-sync.md` Scenario 10 currently specifies:

> Hiding it in Google removes it from the Compass sidebar and stops syncing its events, without touching any other calendar's events or visibility.
> Unhiding it restores it under the same Compass identity — the same visibility preference as before, not reset to a default.

This PR changes that for the `hidden: true && selected: true` case without updating the doc, so the acceptance guide and the code now disagree.

It also doesn't match the reported symptom. If a whole calendar had been inactive, every event on it would be missing, including past ones — but the screenshots show individual events missing on Aug 18–19 while other events from several calendars render normally on Aug 16–17. That's the offset-only timezone bug's signature, not an inactive calendar.

On the premise itself: in Google's model `hidden: true` means the calendar was removed from the list, and its events aren't rendered in the Calendar UI at that point — `selected` is stale rather than authoritative. The test here also asserts `hidden: true` with `selected` omitted stays inactive, so the change only fires on the narrow `hidden && selected` combination. Might be worth confirming against the actual `calendarList` payload for the affected account before shipping it, or splitting it into its own PR so the timezone fix isn't held up.

**2. The events already lost won't come back from this alone.**

The affected connections are healthy with advanced sync cursors, and an incremental pull only returns what changed since the cursor — Google won't re-send an event that was skipped at import time unless the user edits it. `importCalendarEvents` also returns early when `syncCursor !== null` (`calendar-import.service.ts:88`), so the "calendar-list rediscovery enqueues `initialImport`" note in the description only helps calendars that become newly active; it does nothing for offset-only events skipped on calendars that already hold a cursor.

Recovering those needs a full unwindowed re-read — `repairCalendar` already does exactly that into a new generation. I'm adding an operator CLI to enqueue those repairs per account in a follow-up on `claude/compass-calendar-missing-events-y8o29c`, so no action needed here; flagging it mainly so the test plan's "confirm the missing Walk/Haircut events" step doesn't read as expected to pass on this PR alone.

---
_Generated by [Claude Code](https://claude.ai/code)_

tyler-dane:
Correction to point 2 in my comment above — I was wrong, please disregard it.

The reported events have since appeared in Compass on their own, with this PR still an unmerged draft and nothing on `main` touching the import or normalizer path (`4da53b0`, `a3786d2` are unrelated). So they were never blocked by the offset-only `timeZone` throw — if they had been, they could not have shown up without a deploy. My "these won't come back without a repair" claim was built on a diagnosis that turned out not to apply here.

The `unmappableSchedule` throw on offset-only timed events still looks like a real defect worth fixing, and the reader logging is still worth having. I just can't claim it's what caused this report.

Point 1 (the `hidden`/`selected` change vs. acceptance Scenario 10) stands as written.

---
_Generated by [Claude Code](https://claude.ai/code)_

